Issue #60, opened 2020-07-21
Replay training dataset query


SirOibaf:
This allows Hopsworks to store the query object and replay when requested.

PR #73, opened 2020-08-06
#60 Replay training dataset query

SirOibaf:


java/src/main/java/com/logicalclocks/hsfs/TrainingDataset.java
```diff
@@ -157,11 +161,8 @@ public void save(Dataset<Row> dataset) throws FeatureStoreException, IOException
    * @throws IOException
    */
   public void save(Query query, Map<String, String> writeOptions) throws FeatureStoreException, IOException {
-    Dataset<Row> dataset = query.read();
-    trainingDatasetEngine.save(this, dataset, writeOptions);
-    if (statisticsEnabled) {
-      statisticsEngine.computeStatistics(this, dataset);
```
moritzmeister, line 163: why did you remove the statistics computation?

java/src/main/java/com/logicalclocks/hsfs/engine/TrainingDatasetEngine.java
```diff
@@ -65,6 +69,7 @@ public void save(TrainingDataset trainingDataset, Dataset<Row> dataset,
     }
 
     // Update the original object - Hopsworks returns the full location and incremented version
+    trainingDataset.setId(apiTD.getId());
```
moritzmeister, line 72: duplicate of line 75

java/src/main/java/com/logicalclocks/hsfs/engine/Utils.java
```diff
@@ -48,10 +49,22 @@
...
+    int index = 0;
+    for (StructField structField : dataset.schema().fields()) {
+      // TODO(Fabio): unit test this one for complext types
+      features.add(new TrainingDatasetFeature(structField.name(), structField.dataType().catalogString(), index++));
+    }
+
+    return features;
+  }
+
+  public void schemaMatches(Dataset<Row> dataset, List<TrainingDatasetFeature> features) throws FeatureStoreException {
```
moritzmeister, line 64: maybe we can call it `trainingDatasetSchemaMatches` to avoid confusion
```suggestion
  public void trainingDatasetSchemaMatches(Dataset<Row> dataset, List<TrainingDatasetFeature> features) throws FeatureStoreException {
```

python/hsfs/core/training_dataset_engine.py
```diff
@@ -72,6 +72,11 @@ def read(self, training_dataset, split, user_read_options):
             path,
         )
 
+    def query(self, training_dataset, storage):
+        return self._training_dataset_api.get_query(training_dataset)[
+            "queryOnline" if storage == "online" else "query"
```
moritzmeister, line 77: should probably `.lower()` the storage

python/hsfs/feature_group.py
```diff
@@ -67,15 +67,6 @@ def __init__(
         self._default_storage = default_storage
         self._hudi_enabled = hudi_enabled
 
-        if id is None:
```
moritzmeister, line 70: am I stupid, why did I add this twice, thanks for fixing

python/hsfs/training_dataset.py
```diff
@@ -409,3 +419,6 @@ def get_statistics(self, commit_time=None):
             return self.statistics
         else:
             return self._statistics_engine.get(self, commit_time)
+
+    def query(self, storage="online"):
```
moritzmeister, line 423: from an api perspective, I think this could be a property, and then have a separate method `.get_query(storage)` see the statistics, how I did it there. not sure what's better.

moritzmeister (changes requested):
I wasn't able to add a comment there because you didn't touch that part of the file but in `spark.py` engine, I think we need to fix the `schema_matches` function to take into account the order since we use it for inserting data into training datasets. The same as you did in the java part.

java/src/main/java/com/logicalclocks/hsfs/TrainingDataset.java
SirOibaf: No I didn't. `save(Query query... ` calls `save(Dataset<Row>` which has the code to compute the statistics

java/src/main/java/com/logicalclocks/hsfs/engine/TrainingDatasetEngine.java
SirOibaf: done

java/src/main/java/com/logicalclocks/hsfs/engine/Utils.java
SirOibaf: Ok

java/src/main/java/com/logicalclocks/hsfs/TrainingDataset.java
moritzmeister: ah yes, sorry

python/hsfs/core/training_dataset_engine.py
SirOibaf: fixed

python/hsfs/training_dataset.py
SirOibaf: Ok, I see, let's keep it consistent then.

SirOibaf: fixed and added docstring

moritzmeister (approved):
:+1:

PR #76, opened 2020-08-12
[HOPSWORKS-1972] Featurestore connect using API key value

jimdowling:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-1972

Python API Tested against:
https://colab.research.google.com/drive/1ZrOZzBLS4E0ue85sfRwISj5FlL2YiPfV?usp=sharing

Java API - new unit Test added in TestHopsworksExternalClient.java that needs an external cluster (but the unit test does not fail if external cluster not available).

### DIFF
--- a/java/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksExternalClient.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksExternalClient.java
@@ -108,14 +108,14 @@ public class HopsworksExternalClient implements HopsworksHttpClient {
     httpHost = new HttpHost(host, port, "https");
 
     connectionPool = new PoolingHttpClientConnectionManager(
-            createConnectionFactory(httpHost, hostnameVerification, trustStorePath));
+        createConnectionFactory(httpHost, hostnameVerification, trustStorePath));
     connectionPool.setMaxTotal(10);
     connectionPool.setDefaultMaxPerRoute(10);
 
     httpClient = HttpClients.custom()
-            .setConnectionManager(connectionPool)
-            .setKeepAliveStrategy((httpResponse, httpContext) -> 30 * 1000)
-            .build();
+           .setConnectionManager(connectionPool)
+           .setKeepAliveStrategy((httpResponse, httpContext) -> 30 * 1000)
+           .build();
 
     if (!Strings.isNullOrEmpty(apiKeyValue)) {
       this.apiKey = apiKeyValue;
@@ -169,7 +169,6 @@ public class HopsworksExternalClient implements HopsworksHttpClient {
     if (!Strings.isNullOrEmpty(apiKeyFilepath)) {
       return FileUtils.readFileToString(Paths.get(apiKeyFilepath).toFile());
     }
-
 
     switch (secretStore) {
       case PARAMETER_STORE:
--- a/java/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksExternalClient.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksExternalClient.java
@@ -72,30 +72,56 @@ public class HopsworksExternalClient implements HopsworksHttpClient {
 
   private String apiKey = "";
 
-  public HopsworksExternalClient(String host, int port, Region region,
-                                 SecretStore secretStore, boolean hostnameVerification,
-                                 String trustStorePath, String apiKeyFilepath)
-      throws IOException, FeatureStoreException, KeyStoreException, CertificateException,
-      NoSuchAlgorithmException, KeyManagementException {
+  public HopsworksExternalClient(String host, int port, String apiKeyFilepath,
+                                 boolean hostnameVerification, String trustStorePath)
+          throws IOException, FeatureStoreException, KeyStoreException, CertificateException,
+          NoSuchAlgorithmException, KeyManagementException {
+    this(host, port, null, null, hostnameVerification, trustStorePath, apiKeyFilepath, null);
+  }
 
-    httpHost = new HttpHost(host, port, "https");
+  public HopsworksExternalClient(String host, int port, boolean hostnameVerification,
+                                 String trustStorePath, Region region, SecretStore secretStore)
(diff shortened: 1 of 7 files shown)


java/src/test/java/com/logicalclocks/hsfs/TestHopsworksExternalClient.java
```diff
@@ -53,6 +53,21 @@
...
+  @Test
+  public void testReadAPIKey() throws IOException, FeatureStoreException {
+    CloseableHttpClient httpClient = HttpClients.createSystem();
+    try {
+      HopsworksConnection hc = HopsworksConnection.builder().host("35.241.253.100").hostnameVerification(false)
+              .project("demo_featurestore_admin000")
+              .apiKeyValue("ovVQksgJezSckjyK.ftO2YywCI6gZp4btlvWRnSDjSgyAQgCTRAoQTTSXBxPRMo0Dq029eAf3HVq3I6JO").build();
+      System.out.println("Connected");
+      FeatureStore fs = hc.getFeatureStore();
+      Assert.assertTrue(fs != null);
```
moritzmeister, line 65: I am not sure `hc.getFeatureStore()` will ever return `null` but will rather throw an exception if it can't do the rest call. So I don't think this test will ever fail with an AssertionError and hence always run into the catch clause.

python/hsfs/connection.py
```diff
@@ -355,11 +390,20 @@ def cert_folder(self, cert_folder):
...
+    @property
+    def api_key_value(self):
+        return self._api_key_value
+    
     @api_key_file.setter
     @not_connected
     def api_key_file(self, api_key_file):
         self._api_key_file = api_key_file
 
+    @api_key_file.setter
```
moritzmeister, line 402:
```suggestion
    @api_key_value.setter
```
copy paste error

python/hsfs/connection.py
```diff
@@ -355,11 +390,20 @@ def cert_folder(self, cert_folder):
...
+    @not_connected
+    def api_key_value(self, api_key_value):
+        self._api_key_value = api_key_value
+
```
moritzmeister, line 406:
```suggestion

```
remove white space

python/hsfs/connection.py
```diff
@@ -355,11 +390,20 @@ def cert_folder(self, cert_folder):
     def api_key_file(self):
         return self._api_key_file
 
+    @property
+    def api_key_value(self):
+        return self._api_key_value
+
```
moritzmeister, line 396:
```suggestion

```
remove white space

java/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksExternalClient.java
```diff
@@ -72,11 +72,38 @@
...
+    this.httpHost = httpHost;
+  }
+
+  HopsworksExternalClient(String host, int port, Region region,
                                  SecretStore secretStore, boolean hostnameVerification,
-                                 String trustStorePath, String apiKeyFilepath)
-      throws IOException, FeatureStoreException, KeyStoreException, CertificateException,
-      NoSuchAlgorithmException, KeyManagementException {
+                                 String trustStorePath, String apiKeyFilepath, String apiKeyValue)
+          throws IOException, FeatureStoreException, KeyStoreException, CertificateException,
```
moritzmeister, line 105: it's strange that the checkstyle-plugin doesn't complain about the additional white space before the `throws`. but ok, then it should be fine

java/src/test/java/com/logicalclocks/hsfs/TestHopsworksExternalClient.java
moritzmeister: I think it is an IOException that it is throwing. Fabio wants to change it to a FeatureStoreException

python/hsfs/connection.py
jimdowling: Fixed.

python/hsfs/connection.py
jimdowling: Done.

java/src/test/java/com/logicalclocks/hsfs/TestHopsworksExternalClient.java
```diff
@@ -53,6 +53,21 @@
         .willReturn(success().body(HttpBodyConverter.json(credentials)))
     ));
 
+  @Test
+  public void testReadAPIKey() throws IOException, FeatureStoreException {
```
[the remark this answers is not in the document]
jimdowling: The options I see are:
1. comment it out and leave the code there, so it can easily be tested again
2. delete the code.

What do you think?

java/src/test/java/com/logicalclocks/hsfs/TestHopsworksExternalClient.java
jimdowling: The code is removed.

python/hsfs/connection.py
jimdowling: Done

java/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksExternalClient.java
jimdowling: If checkstyle doesn't complain, that is good enough for me. Is that ok?

moritzmeister: yes as long as checkstyle isn't complaining, it's fine

moritzmeister:
@jimdowling I don't see any new commits, did you push them?

java/src/test/java/com/logicalclocks/hsfs/TestHopsworksExternalClient.java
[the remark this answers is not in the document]
moritzmeister: We can leave it commented out, but I still believe the `hc.getFeatureStore()` will never return null but instead throw a exception if it can't get the feature store reference and hence the test will never get to the assert and always pass as the exception gets caught by the catch statement.

moritzmeister (approved):

PR #79, opened 2020-08-16
Pydoop should handle hopsfs:// scheme

### LINKED ISSUE #78: Pydoop should handle hopsfs:// scheme
Currently there is an issue related to the scheme splitting. Monkey patch the class so that it handles `hopsfs://` scheme as well.

SirOibaf:
Closes #78

@moritzmeister - if you have a better place to locate this piece of code, let me know

### DIFF
--- a/python/hsfs/engine/spark.py
+++ b/python/hsfs/engine/spark.py
@@ -14,10 +14,10 @@
 #   limitations under the License.
 #
 
+import os
+
 import pandas as pd
 import numpy as np
-
-import os
 
 # in case importing in %%local
 try:
--- a/python/hsfs/engine/spark.py
+++ b/python/hsfs/engine/spark.py
@@ -17,6 +17,8 @@
 import pandas as pd
 import numpy as np
 
+import os
+
 # in case importing in %%local
 try:
     from pyspark.sql import SparkSession, DataFrame
--- a/python/hsfs/engine/spark.py
+++ b/python/hsfs/engine/spark.py
@@ -40,7 +40,9 @@ class Engine:
         self._spark_session.conf.set("hive.exec.dynamic.partition", "true")
         self._spark_session.conf.set("hive.exec.dynamic.partition.mode", "nonstrict")
 
-        self._setup_pydoop()
+        if not os.path.exists("/dbfs/"):
+            # If we are on Databricks don't setup Pydoop as it's not available and cannot be easily installed.
+            self._setup_pydoop()
 
     def sql(self, sql_query, feature_store, online_conn, dataframe_type):
         if not online_conn:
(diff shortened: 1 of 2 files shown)


python/hsfs/engine/spark.py
```diff
@@ -275,3 +275,34 @@ def _setup_s3(self, storage_connector, path):
 
 class SchemaError(Exception):
     """Thrown when schemas don't match"""
+
+
+# Replace Pydoop split method to be able to support hopsfs:// schemes
+class _HopsFSPathSplitter(hdfs.path._HdfsPathSplitter):
```
moritzmeister, line 281: are we missing the import of hdfs? or where does it come from?

python/hsfs/engine/spark.py
```diff
@@ -275,3 +275,34 @@ def _setup_s3(self, storage_connector, path):
...
+    import os
```
moritzmeister, line 282: are these imports actually needed? if yes, can they be moved to the top level?

python/hsfs/engine/spark.py
```diff
@@ -275,3 +275,34 @@ def _setup_s3(self, storage_connector, path):
...
+            if not path.startswith("/"):
+                path = "/user/%s/%s" % (user, path)
+        elif scheme == "file":
+            hostname, port, path = "", 0, netloc + path
+        else:
+            cls.raise_bad_path(hdfs_path, "unsupported scheme %r" % scheme)
+        return hostname, port, path
+
+
+hdfs.path._HdfsPathSplitter = _HopsFSPathSplitter
```
moritzmeister, line 308: Did you test this? We are always importing this module in the __init__.py of the engine package, so I would think it will fail on SageMaker. I think the class is fine to have in this module, but we should move this line to be executed only when we have pydoop available. Or we make pydoop a dependency so it's also available on other environments.

python/hsfs/engine/spark.py
SirOibaf: It doesn't look like, I think it was just a left over from the Jupyter notebook I was testing it

python/hsfs/engine/spark.py
SirOibaf: I see your point. No, we can't make Pydoop available everywhere, as installing Pydoop requires having an Hadoop distribution available. 

Maybe I can move the initialization in the constructor so it runs only if the Spark engine is actually created.

moritzmeister: yes either that, or moving it to this if-clause here, https://github.com/logicalclocks/feature-store-api/blob/eb1312283d12ea9e0df1a864187df2c3678a168c/python/hsfs/engine/__init__.py#L26

python/hsfs/engine/spark.py
```diff
@@ -17,6 +17,8 @@
 import pandas as pd
 import numpy as np
 
+import os
```
moritzmeister, line 20: nitpick, standard library imports go to the top

SirOibaf: Fixed

PR #84, opened 2020-08-18
[HOPSWORKS-1982] Deequ statistics for Feature Groups/Training Datasets

moritzmeister:


python/hsfs/core/feature_group_api.py
```diff
@@ -112,3 +112,15 @@ def delete(self, feature_group_instance):
...
+        """Update the metadata of a feature group.
+
+        It is only possible to update the description and statistics config
+        of a feature group.
+
+        :param feature_group_instance: [description]
+        :type feature_group_instance: [type]
+        :raises NotImplementedError: [description]
+        """
+        raise NotImplementedError
```
SirOibaf, line 126: Reminder.

python/hsfs/feature_group.py
```diff
@@ -154,9 +160,26 @@ def insert(self, features, overwrite=False, storage=None, write_options={}):
...
 
+        if self.statistics_config.enabled:
+            self._statistics_engine.compute_statistics(self, self.read(storage))
+
     def delete(self):
         self._feature_group_engine.delete(self)
 
+    def update(self):
+        self._feature_group_engine.update(self)
+        return self
```
SirOibaf, line 171: What is this method supposed to do?

python/hsfs/feature_group.py
```diff
@@ -154,9 +160,26 @@ def insert(self, features, overwrite=False, storage=None, write_options={}):
             write_options,
         )
 
+        if self.statistics_config.enabled:
+            self._statistics_engine.compute_statistics(self, self.read(storage))
```
SirOibaf, line 164: nitpick - we can call the compute_statistics method here.

python/hsfs/feature_group.py
```diff
@@ -154,9 +160,26 @@ def insert(self, features, overwrite=False, storage=None, write_options={}):
...
+
+    def compute_statistics(self, storage=None):
+        """Recompute the statistics for the feature group and save them to the
+        feature store.
+
+        :param storage: storage to read data from, defaults to None
+        :type storage: str, optional
+        """
+        if self.statistics_config.enabled:
+            self._statistics_engine.compute_statistics(self, self.read(storage))
```
SirOibaf, line 181: `storage` here can be `ALL` meaning that the insert will be written both in the online and offline feature store.
Read will throw an exception if `ALL` is set, so the user will have an exception during the insert.

Also I don't think it's wise to let the user compute the statistics over the online feature store. It will trigger a scan which is not great for NDB.

python/hsfs/training_dataset.py
```diff
@@ -108,8 +114,13 @@ def save(self, features, write_options={}):
             )
 
         user_version = self._version
+        user_stats_config = self._statistics_config
         self._features = engine.get_instance().parse_schema(feature_dataframe)
         self._training_dataset_engine.save(self, feature_dataframe, write_options)
+        # currently we do not save the training dataset statistics config for training datasets
+        self.statistics_config = user_stats_config
```
SirOibaf, line 121: Why are you doing this? (assigning to a different variable?)

python/hsfs/training_dataset.py
```diff
@@ -130,9 +141,19 @@ def insert(self, features, overwrite, write_options={}):
             self, feature_dataframe, write_options, overwrite
         )
 
+        if self.statistics_config.enabled:
+            self._statistics_engine.compute_statistics(self, self.read())
```
SirOibaf, line 145: call the `compute_statistics` method.

python/hsfs/core/statistics_engine.py
```diff
@@ -0,0 +1,39 @@
...
+
+class StatisticsEngine:
+    def __init__(self, feature_store_id, entity_type):
+        self._statistics_api = statistics_api.StatisticsApi(
+            feature_store_id, entity_type
+        )
+
+    def compute_statistics(self, metadata_instance, feature_dataframe):
+        """Compute statistics for a dataframe and send the result json to Hopsworks."""
+        commit_str = datetime.datetime.now().strftime("%Y%m%d%H%M%S")
```
SirOibaf, line 31: I think for the moment it's ok, but as when we port Hudi, the commit id should come as parameter to the method call.

python/hsfs/core/statistics_api.py
```diff
@@ -0,0 +1,46 @@
...
+
+        :param feature_store_id: id of the respective featurestore
+        :type feature_store_id: int
+        :param entity_type: "trainingdatasets" or "featuregroups"
+        :type entity_type: str
+        """
+        self._feature_store_id = feature_store_id
+        self._entity_type = entity_type
+
+    def post(self, metadata_instance, statistics):
```
SirOibaf, line 32: I think we should also allow users to get the statistics from the API.

java/pom.xml
```diff
@@ -29,6 +30,12 @@
             <version>${lombok.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>com.amazon.deequ</groupId>
+            <artifactId>deequ</artifactId>
+            <version>${deequ.version}</version>
```
SirOibaf, line 36: We need to change this. Otherwise we won't get the changes we made.

python/hsfs/feature_group.py
moritzmeister: similarly to `.save()` and `.insert()` this is the method that triggers the update of metadata. It is only possible to update the description and statistics config of a feature group. The user can change these on the object and then call `.update()` to persist them.

python/hsfs/feature_group.py
moritzmeister: okay if storage is `ALL` I can just compute it on the offline FS, but do we really want to disable statistics completely for the online feature store?

python/hsfs/training_dataset.py
moritzmeister: for training dataset we aren't saving the config in the database along the metadata. the `engine.save()` before will update `self` with the metadata from the backend, hence overwriting the config. I am saving it in this temporary variable `user_stats_config` so I can set it again afterwards

python/hsfs/core/statistics_engine.py
moritzmeister: totally agree, commit id should be generated by hudi and passed here. I made sure it's the same format for now.

python/hsfs/feature_group.py
moritzmeister: Removed the possibility to specify a storage to compute statistics from and changed it so it issues a warning if the default storage is `online`. And read from offline if default storage is `ALL`.

python/hsfs/feature_group.py
moritzmeister: done

python/hsfs/training_dataset.py
moritzmeister: done

java/pom.xml
moritzmeister: this requires building deequ from our branch before we build hsfs. not sure what needs to be done for this.

python/hsfs/core/statistics_api.py
moritzmeister: done

moritzmeister: done

python/hsfs/core/feature_group_api.py
moritzmeister: addressed, now you're able to update only the statistics config

java/pom.xml
SirOibaf: I think there is already a pipeline to build deequ. I can take care of fixing the group name.

python/hsfs/feature_group.py
SirOibaf: Ok

python/hsfs/feature_group.py
```diff
@@ -284,3 +333,43 @@ def partition_key(self, new_partition_key):
...
+
+        :param commit_time: Commit time in the format `YYYYMMDDhhmmss`, defaults to None
+        :type commit_time: str, optional
+        :return: Statistics information
+        :rtype: Statistics
+        """
+        if commit_time is None:
+            return self.statistics
+        else:
+            return self._statistics_engine.get(self, commit_time)
```
SirOibaf, line 375: I'd return the statistics content here, rather than the statistics object. Or at least the `to_dict()`. I doubt people are interested in the remaining fields.

python/hsfs/training_dataset.py
SirOibaf: ok

python/hsfs/training_dataset.py
```diff
@@ -351,3 +371,43 @@ def seed(self):
...
+
+        :param commit_time: Commit time in the format `YYYYMMDDhhmmss`, defaults to None
+        :type commit_time: str, optional
+        :return: Statistics information
+        :rtype: Statistics
+        """
+        if commit_time is None:
+            return self.statistics
+        else:
+            return self._statistics_engine.get(self, commit_time)
```
SirOibaf, line 413: Same as for the feature groups.

SirOibaf (changes requested):

python/hsfs/feature_group.py
moritzmeister: I am returning a `Statistics` object here, the same way as I do in the Java part. The statistics object only has the other fields like href as private members, and the object has no properties/getters for those fields. I can also get rid of the fields entirely.

java/pom.xml
SirOibaf: I published the our fork of deequ using the `com.logicalclocks` group, so that is easier to manage: https://archiva.hops.works/#artifact/com.logicalclocks/deequ

python/hsfs/feature_group.py
SirOibaf: Ok

SirOibaf (approved):
LGTM, Just change the group id of deequ and fix the conflict.

moritzmeister:
Will rebase after the Java part was merged.

PR #111, opened 2020-10-05
[HOPSWORKS-1580] Change endpoint for getting the online storage connector

gibchikafa:


java/pom.xml
```diff
@@ -199,6 +200,15 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
```
moritzmeister, line 203: why do you need to add this?

python/hsfs/core/storage_connector_api.py
```diff
@@ -18,6 +18,7 @@
 
 
 class StorageConnectorApi:
+    CONST_ONLINE_FEATURE_STORE_CONNECTOR_SUFFIX = "_onlinefeaturestore"
```
moritzmeister, line 21:
```suggestion
    CONST_ONLINE_FEATURE_STORE_CONNECTOR_SUFFIX = "_onlinefeaturestore"
    
```
add a new line

moritzmeister (changes requested):
looks good, one minor style change please

java/src/main/java/com/logicalclocks/hsfs/metadata/StorageConnectorApi.java
```diff
@@ -63,15 +64,17 @@ public StorageConnector getOnlineStorageConnector(FeatureStore featureStore)
...
 
     String uri = UriTemplate.fromTemplate(pathTemplate)
         .set("projectId", featureStore.getProjectId())
         .set("fsId", featureStore.getId())
         .expand();
 
     LOGGER.info("Sending metadata request: " + uri);
-    return hopsworksClient.handleRequest(new HttpGet(uri), StorageConnector.class);
+    StorageConnector[] storageConnectors =  hopsworksClient.handleRequest(new HttpGet(uri), StorageConnector[].class);
+    return Arrays.stream(storageConnectors).filter(s -> s.getName().contains(ONLINE_FEATURE_STORE_CONNECTOR_SUFFIX))
```
moritzmeister, line 76: can you please double check if the backend lower cases the connector name before sending it? or if we even allow mixed case storage connector names.

java/pom.xml
gibchikafa: Oh my bad...I couldnt compile due to the java version. I checked it out to master

java/src/main/java/com/logicalclocks/hsfs/metadata/StorageConnectorApi.java
gibchikafa: The online connector name has a specific format. It is project_name + username + ONLINE_FEATURE_STORE_CONNECTOR_SUFFIX. 
In this PR for backend we created the following query @NamedQuery(name = "FeaturestoreJdbcConnector.findByName", query = "SELECT fsjdbc FROM FeaturestoreJdbcConnector " +
        "fsjdbc WHERE lower(fsjdbc.name)  = lower(:name)")}) so that they are unique regardless of case

moritzmeister: yes sorry, I misread the code, I thought you are filtering by the entire name, not just the suffix.

moritzmeister (approved):

PR #117, opened 2020-10-14
[HOPSWORKS-2009] Append Features to Feature Group schema

moritzmeister:
usage:

```scala
var houses_sold = fs.getFeatureGroup("houses_sold", 1)
val l = Seq(new Feature("asdgadfg", "double", "5.0"))
houses_sold.appendFeatures(l)
```
or
```scala
houses_sold.appendFeatures(new Feature("asdgadfg", "double", "5.0"))
```

### DIFF
--- a/java/src/main/java/com/logicalclocks/hsfs/FeatureGroup.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/FeatureGroup.java
@@ -23,10 +23,12 @@ import com.logicalclocks.hsfs.engine.FeatureGroupEngine;
 import com.logicalclocks.hsfs.engine.StatisticsEngine;
 import com.logicalclocks.hsfs.metadata.Query;
 import com.logicalclocks.hsfs.metadata.Statistics;
+import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Getter;
 import lombok.NonNull;
 import lombok.Setter;
+import lombok.With;
 import org.apache.spark.sql.Dataset;
 import org.apache.spark.sql.Row;
 import org.apache.spark.sql.SaveMode;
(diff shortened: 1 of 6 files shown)


SirOibaf (approved):
I think it's ok for the moment. If in the future we see that we are adding too many of them, we can refactor and deprecate these methods.

java/src/main/java/com/logicalclocks/hsfs/FeatureGroup.java
```diff
@@ -211,6 +217,43 @@ public void updateStatisticsConfig() throws FeatureStoreException, IOException {
...
+   * Append a single feature to the schema of the feature group.
+   * It is only possible to append features to a feature group. Removing features is considered a breaking change.
+   *
+   * @param features
+   * @throws FeatureStoreException
+   * @throws IOException
+   */
+  public void appendFeatures(Feature features) throws FeatureStoreException, IOException {
+    List<Feature> featureList = new ArrayList<>();
+    featureList.add(features);
```
SirOibaf, line 253: You can use `Arrays.asList()` here.

java/src/main/java/com/logicalclocks/hsfs/FeatureGroup.java
```diff
@@ -211,6 +217,43 @@ public void updateStatisticsConfig() throws FeatureStoreException, IOException {
...
+  /**
+   * Append features to the schema of the feature group.
+   * It is only possible to append features to a feature group. Removing features is considered a breaking change.
+   *
+   * @param features
+   * @throws FeatureStoreException
+   * @throws IOException
+   */
+  public void appendFeatures(List<Feature> features) throws FeatureStoreException, IOException {
+    featureGroupEngine.appendFeatures(this, new ArrayList<>(features));
```
SirOibaf, line 240: Why the `ArrayList<>` here? You already have a list.

moritzmeister: that list is immutable for some reason or didn't have the `addAll` method

PR #119, opened 2020-10-16
[HOPSWORKS-2084] Remove default_storage option from feature groups

moritzmeister:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-2084

### DIFF
--- a/java/src/main/java/com/logicalclocks/hsfs/Storage.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/Storage.java
@@ -18,6 +18,5 @@ package com.logicalclocks.hsfs;
 
 public enum Storage {
   OFFLINE,
-  ONLINE,
-  ALL
+  ONLINE
 }
--- a/java/src/main/java/com/logicalclocks/hsfs/metadata/Query.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/metadata/Query.java
@@ -147,40 +147,36 @@ public class Query {
   }
 
   public Dataset<Row> read() throws FeatureStoreException, IOException {
-    return read(Storage.OFFLINE, null);
+    return read(false, null);
   }
 
-  public Dataset<Row> read(Storage storage, Map<String,String> readOptions)
-          throws FeatureStoreException, IOException {
-    if (storage == null) {
-      throw new FeatureStoreException("Storage not supported");
-    }
+  public Dataset<Row> read(boolean online) throws FeatureStoreException, IOException {
+    return read(online, null);
+  }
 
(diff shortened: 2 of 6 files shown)


java/src/main/java/com/logicalclocks/hsfs/metadata/Query.java
```diff
@@ -106,36 +106,33 @@ public Query joinFeatures(Query subquery, List<Feature> leftOn, List<Feature> ri
...
-            = storageConnectorApi.getOnlineStorageConnector(leftFeatureGroup.getFeatureStore());
-        return SparkEngine.getInstance().jdbc(onlineConnector, sqlQuery);
-      default:
-        throw new FeatureStoreException("Storage not supported");
+    return read(false);
+  }
+
+  public Dataset<Row> read(boolean online) throws FeatureStoreException, IOException {
+    String sqlQuery;
+    if (!online) {
```
SirOibaf, line 114: please invert the branches and remove the `!`

java/src/main/java/com/logicalclocks/hsfs/Storage.java
```diff
@@ -18,6 +18,5 @@
 
 public enum Storage {
   OFFLINE,
-  ONLINE,
-  ALL
```
SirOibaf, line 22: why removing `ALL` - assuming you want to insert on both online and offline do you have to call the method twice? Do we want that?

SirOibaf (changes requested):

java/src/main/java/com/logicalclocks/hsfs/Storage.java
moritzmeister: as we discussed, the logic is now the following:

- we get rid of the default storage, but we keep the storage option in the read and insert methods
- save: saves offline, both if online is enabled
- read: defaults to reading from offline, unless you pass storage online
- insert: same as save, but you can overwrite by passing the storage
- user wants to write only to the online feature store: create empty FG with online enabled, use insert.

so if you don't specify Storage, it will insert in both if online_enabled. if you specify Storage, either Online or offline, it will only insert in one of the two.

moritzmeister:
@davitbzh can you please review this, I rebased on your Hudi changes, but I am not entirely sure in which cases e.g. online/offline feature group what the arguments related to hudi should be.

https://logicalclocks.atlassian.net/browse/HOPSWORKS-2084
here's the description of the desired behaviour

davitbzh (approved):

SirOibaf (approved):

PR #120, opened 2020-10-18
[HOPSWORKS-1374] Add Hudi API to HSFS and commits to the FeatureStore metadata

davitbzh:


java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
```diff
@@ -0,0 +1,293 @@
...
+    Seq<String> partitionColumns = utils.getPartitionColumns(featureGroup);
+
+    if (!partitionColumns.isEmpty()) {
+      hudiArgs.put(Constants.HUDI_PARTITION_FIELD, partitionColumns.mkString(","));
+      // For precombine key take 1st particion key
+      hudiArgs.put(Constants.HUDI_PRECOMBINE_FIELD, partitionColumns.head());
+      hudiArgs.put(Constants.HUDI_HIVE_SYNC_PARTITION_FIELDS, partitionColumns.mkString(","));
+    }
+
+    String tableName = featureGroup.getName() + "_" + featureGroup.getVersion();
```
SirOibaf, line 205: We already have a method in `Util.java` that does the same thing. We should that one instead.

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
```diff
@@ -0,0 +1,293 @@
...
+    hudiArgs.put(Constants.HUDI_HIVE_SYNC_ENABLE, "true");
+    hudiArgs.put(Constants.HUDI_HIVE_SYNC_TABLE, tableName);
+    // Gets the JDBC Connector for the project's Hive metastore
+    StorageConnectorApi storageConnectorApi = new StorageConnectorApi();
+    StorageConnector storageConnector = storageConnectorApi.getByNameAndType(featureGroup.getFeatureStore(),
+                featureGroup.getFeatureStore().getName(), StorageConnectorType.JDBC);
+    String connStr = storageConnector.getConnectionString();
+    String pw = FileUtils.readFileToString(new File("material_passwd"));
+    String jdbcUrl = connStr + "sslTrustStore=t_certificate;trustStorePassword="
+                + pw + ";sslKeyStore=k_certificate;keyStorePassword=" + pw;
```
SirOibaf, line 219: I think this should be moved either in `Utils.java` or in the `StorageConnector` class itself.

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
```diff
@@ -0,0 +1,293 @@
...
+    hudiArgs.put(Constants.HUDI_HIVE_SYNC_JDBC_URL, jdbcUrl);
+    hudiArgs.put(Constants.HUDI_HIVE_SYNC_DB, featureGroup.getFeatureStore().getName());
+
```
SirOibaf, line 222: Users are allowed to pass a Map with configuration options when calling inserts. Here we should put all the elements from that map in the `hudiArgs`. (`hudiArgs.putAll()`) in this way users can have full control over the framework if they need to or if there is a bug.

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
```diff
@@ -0,0 +1,293 @@
...
+    long totalRecordsDeleted = commitMetadata.getTotalRecordsDeleted();
+    fgCommitMetadata.setRowsDeleted(totalRecordsDeleted);
+    return fgCommitMetadata;
+  }
+
+  @SneakyThrows
+  private String timeStampToHudiFormat(Long commitedOnTimeStamp) {
+    Date commitedOnDate = new Timestamp(commitedOnTimeStamp);
+    String strCommitedOnDate = dateFormat.format(commitedOnDate);
+    return strCommitedOnDate;
```
SirOibaf, line 291:
```suggestion
    return dateFormat.format(commitedOnDate);
```

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
```diff
@@ -0,0 +1,293 @@
...
+
+    fgCommitMetadata.setCommitDateString(commitTimeline.lastInstant().get().getTimestamp());
+    byte[] commitsToReturn = commitTimeline.getInstantDetails(commitTimeline.lastInstant().get()).get();
+    HoodieCommitMetadata commitMetadata = HoodieCommitMetadata.fromBytes(commitsToReturn,HoodieCommitMetadata.class);
+    long totalUpdateRecordsWritten = commitMetadata.fetchTotalUpdateRecordsWritten();
+    fgCommitMetadata.setRowsUpdated(totalUpdateRecordsWritten);
+    long totalInsertRecordsWritten = commitMetadata.fetchTotalInsertRecordsWritten();
+    fgCommitMetadata.setRowsInserted(totalInsertRecordsWritten);
+    long totalRecordsDeleted = commitMetadata.getTotalRecordsDeleted();
+    fgCommitMetadata.setRowsDeleted(totalRecordsDeleted);
```
SirOibaf, line 283:
```suggestion
    fgCommitMetadata.setRowsUpdated(commitMetadata.fetchTotalUpdateRecordsWritten());
    fgCommitMetadata.setRowsInserted(commitMetadata.fetchTotalInsertRecordsWritten());
    fgCommitMetadata.setRowsDeleted(commitMetadata.getTotalRecordsDeleted());
```

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
```diff
@@ -0,0 +1,293 @@
...
+    return hudiArgs;
+  }
+
+  //hudi time time travel sql query
+  public void registerTemporaryTable(SparkSession sparkSession, FeatureGroup featureGroup, String alias,
+                                     Long startTimestamp, Long  endTimestamp) {
+
+    sparkSession.conf().set("spark.sql.hive.convertMetastoreParquet", "false");
+    sparkSession.sparkContext().hadoopConfiguration().setClass("mapreduce.input.pathFilter.class",
+        org.apache.hudi.hadoop.HoodieROTablePathFilter.class, org.apache.hadoop.fs.PathFilter.class);
```
SirOibaf, line 232: Can't we set these globally when we instantiate the SparkEngine?

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
```diff
@@ -0,0 +1,293 @@
...
+      hudiArgs.put(Constants.HUDI_QUERY_TYPE_OPT_KEY, Constants.HUDI_QUERY_TYPE_SNAPSHOT_OPT_VAL);
+      basePath =  basePath + StringUtils.repeat("/*", numberOfpartitionCols + 1);
+    } else if (endTimestamp != null) {
+      hudiArgs.put(Constants.HUDI_QUERY_TYPE_OPT_KEY, Constants.HUDI_QUERY_TYPE_INCREMENTAL_OPT_VAL);
+      String hudiCommitEndTime = timeStampToHudiFormat(endTimestamp);
+      //point in time  query
+      if (startTimestamp == null) {
+        hudiArgs.put(Constants.HUDI_BEGIN_INSTANTTIME_OPT_KEY, "000");
+        hudiArgs.put(Constants.HUDI_END_INSTANTTIME_OPT_KEY, hudiCommitEndTime);
+      } else if (startTimestamp != null) {       //incremental  query
```
SirOibaf, line 257:
```suggestion
      } else {       //incremental  query
```
`startTimestamp != null` is always true if we reach this point.

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
```diff
@@ -0,0 +1,293 @@
...
+        String hudiCommitStartTime = timeStampToHudiFormat(startTimestamp);
+        hudiArgs.put(Constants.HUDI_BEGIN_INSTANTTIME_OPT_KEY, hudiCommitStartTime);
+        hudiArgs.put(Constants.HUDI_END_INSTANTTIME_OPT_KEY, hudiCommitEndTime);
+      }
+    }
+
```
SirOibaf, line 263: Same for the write arguments. Users have the possibility of passing a Map<String, String> with options to use in the read. We should put all the user provided options in the `hudiArgs` map.

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
```diff
@@ -0,0 +1,293 @@
...
+  private Map<String, String> setupHudiReadOpts(FeatureGroup featureGroup, Long startTimestamp, Long endTimestamp) {
+    Map<String, String> hudiArgs = new HashMap<String, String>();
+    Integer numberOfpartitionCols = utils.getPartitionColumns(featureGroup).length();
+    String basePath = featureGroup.getLocation();
+
+    //Snapshot query
+    if (startTimestamp == null && endTimestamp == null) {
+      hudiArgs.put(Constants.HUDI_QUERY_TYPE_OPT_KEY, Constants.HUDI_QUERY_TYPE_SNAPSHOT_OPT_VAL);
+      basePath =  basePath + StringUtils.repeat("/*", numberOfpartitionCols + 1);
```
SirOibaf, line 249: `basePath` is not used.

java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java
```diff
@@ -290,15 +302,29 @@ public void writeOnlineDataframe(Dataset<Row> dataset, SaveMode saveMode, Map<St
...
+      hudiEngine.saveHudiFeatureGroup(sparkSession,featureGroup, dataset, saveMode, operation);
+    } else {
+      writeSparkDataset(featureGroup, dataset, saveMode,  writeOptions);
+    }
+
+  }
+
+  private void writeSparkDataset(FeatureGroup featureGroup, Dataset<Row> dataset,
+                                 SaveMode saveMode,  Map<String, String> writeOptions) {
+
```
SirOibaf, line 318: please don't add random empty spaces.

java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java
```diff
@@ -329,4 +355,10 @@ public String profile(Dataset<Row> df, boolean correlation, boolean histogram) {
   public String profile(Dataset<Row> df) {
     return profile(df, null, true, true);
   }
+
+  public FeatureGroupCommit commitDelete(FeatureGroup featureGroup, Dataset<Row> deleteDF)
+      throws IOException, FeatureStoreException {
+    FeatureGroupCommit featureGroupCommit = hudiEngine.deleteRecord(sparkSession, featureGroup, deleteDF);
+    return featureGroupCommit;
```
SirOibaf, line 362:
```suggestion
    return hudiEngine.deleteRecord(sparkSession, featureGroup, deleteDF);
```

java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java
```diff
@@ -329,4 +355,10 @@ public String profile(Dataset<Row> df, boolean correlation, boolean histogram) {
   public String profile(Dataset<Row> df) {
     return profile(df, null, true, true);
   }
+
+  public FeatureGroupCommit commitDelete(FeatureGroup featureGroup, Dataset<Row> deleteDF)
```
SirOibaf, line 359: This method should be in the `FeatureGroupEngine` not here

java/src/main/java/com/logicalclocks/hsfs/metadata/FeatureGroupApi.java
```diff
@@ -148,4 +151,46 @@ public FeatureGroup updateMetadata(FeatureGroup featureGroup, String queryParame
...
+        .set("projectId", featureGroupBase.getFeatureStore().getProjectId())
+        .set("fsId", featureGroupBase.getFeatureStore().getId())
+        .set("fgId", featureGroupBase.getId())
+        .set("limit", limit)
+        .expand();
+
+    LOGGER.info("Sending metadata request: " + uri);
+    FeatureGroupCommit[] commitDetails = hopsworksClient.handleRequest(new HttpGet(uri), FeatureGroupCommit[].class);
+
+    return commitDetails;
```
SirOibaf, line 194:
```suggestion
    return hopsworksClient.handleRequest(new HttpGet(uri), FeatureGroupCommit[].class);
```

java/src/main/java/com/logicalclocks/hsfs/engine/FeatureGroupEngine.java
```diff
@@ -56,9 +60,8 @@
    * @throws FeatureStoreException
    * @throws IOException
    */
-  public void saveFeatureGroup(FeatureGroup featureGroup, Dataset<Row> dataset,
-                               List<String> primaryKeys, List<String> partitionKeys,
-                               Storage storage, Map<String, String> writeOptions)
+  public void saveFeatureGroup(FeatureGroup featureGroup, Dataset<Row> dataset, List<String> primaryKeys,
```
SirOibaf, line 63: What happened to the `precombine` flat of the feature object? why don't we set it here?

java/src/main/java/com/logicalclocks/hsfs/metadata/Query.java
```diff
@@ -43,6 +46,10 @@
   private FeatureGroup leftFeatureGroup;
   @Getter @Setter
   private List<Feature> leftFeatures;
+  @Getter @Setter
+  private String leftFeaturegroupStartTime;
```
SirOibaf, line 50: `FeatureGroup` consistency.

SirOibaf (changes requested):
Reviewed Java part

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
davitbzh: I think this is only needed for Hudi, for this particular function. Otherwise we need to check if FG is hudi type and set it globally in SparkEngine

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
davitbzh: done. Currently I am using the same writeOptions MAP that can be used also for SPARK. User just need to populate it accordingly

java/src/main/java/com/logicalclocks/hsfs/engine/FeatureGroupEngine.java
davitbzh: For now I decided to omit as you mention that needed to be discussed. But adding will not be problem

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
SirOibaf: These configurations are global anyway. They will be applied globally, but only after you have read the first Hudi feature group. I'd say we are consistent and setup Spark properly when starting.

java/src/main/java/com/logicalclocks/hsfs/engine/FeatureGroupEngine.java
SirOibaf: ok

python/hsfs/core/hudi_engine.py
```diff
@@ -0,0 +1,195 @@
...
+        self._spark_context = spark_context
+        self._spark_session = spark_session
+        self._feature_store_id = feature_store_id
+        self._feature_store_name = feature_store_name
+        self._base_path = self._feature_group.location
+        self._table_name = feature_group.name + "_" + str(feature_group.version)
+
+        # TODO (Davit): Need to fix ComplexKeyGenerator bug
+        self._primary_key = feature_group.primary_key[0]
+        self._partition_key = feature_group.partition_key[0]
```
SirOibaf, line 78: Why are you considering only the first partition/primary key? In Java you build a comma separated list, I guess you should do the same here.

python/hsfs/core/hudi_engine.py
```diff
@@ -0,0 +1,195 @@
...
+        self._spark_session.conf.set("spark.sql.hive.convertMetastoreParquet", "false")
+
+        hudi_options = self._setup_hudi_read_opts(start_timestamp, end_timestamp)
+        self._spark_session.read.format(self.HUDI_SPARK_FORMAT).options(
+            **hudi_options
+        ).load(self._base_path).registerTempTable(alias)
+
+    def _write_hudi_dataset(self, dataset, save_mode, operation, write_options):
+        hudi_options = self._setup_hudi_write_opts(operation)
+        if bool(write_options):
```
SirOibaf, line 106: empty dictionaries evaluate to false in python. So `if write_options` would have been enough. Again, as In java, I would prefer if the options were added in the `_setup_hudi_write_opts` method, so no matter how many times we call it, we know that we need to provide the overwrite parameters.

python/hsfs/core/hudi_engine.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def _get_last_commit_metadata(self, spark_context, base_path):
+        hopsfs_conf = spark_context._jvm.org.apache.hadoop.fs.FileSystem.get(
+            spark_context._jsc.hadoopConfiguration()
+        )
+        commit_timeline = spark_context._jvm.org.apache.hudi.HoodieDataSourceHelpers.allCompletedCommitsCompactions(
+            hopsfs_conf, base_path
+        )
+
+        commits2return = commit_timeline.getInstantDetails(
```
SirOibaf, line 156: Let's try to be consistent and use snake case everywhere.

python/hsfs/core/hudi_engine.py
```diff
@@ -0,0 +1,195 @@
...
+        )
+        return feature_group_commit.FeatureGroupCommit(
+            commitid=None,
+            commit_date_string=commit_timeline.lastInstant().get().getTimestamp(),
+            rows_inserted=commit_metadata.fetchTotalInsertRecordsWritten(),
+            rows_updated=commit_metadata.fetchTotalUpdateRecordsWritten(),
+            rows_deleted=commit_metadata.getTotalRecordsDeleted(),
+        )
+
+    def _timestamp2hudiformat(self, timestamp):
```
SirOibaf, line 171: snake case.

python/hsfs/core/query.py
```diff
@@ -22,27 +22,44 @@
 
 class Query:
     def __init__(
-        self, feature_store_name, feature_store_id, left_feature_group, left_features
+        self,
+        feature_store_name,
+        feature_store_id,
+        left_feature_group,
+        left_features,
+        left_featuregroup_starttime=None,
```
SirOibaf, line 30: snake case for `starttime`

python/hsfs/core/query.py
```diff
@@ -22,27 +22,44 @@
...
+        left_featuregroup_endtime=None,
```
SirOibaf, line 31: snake case for endtime

python/hsfs/core/query.py
```diff
@@ -22,27 +22,44 @@
...
         self._joins = []
         self._query_constructor_api = query_constructor_api.QueryConstructorApi()
         self._storage_connector_api = storage_connector_api.StorageConnectorApi(
             feature_store_id
         )
 
-    def read(self, storage="offline", dataframe_type="default"):
+    def read(
+        self, storage="offline", dataframe_type="default", time_travel_fomat="HUDI",
+    ):
```
SirOibaf, line 47: Why do you need the `time_travel_format` option here? That should come from the feature group involved in the join

SirOibaf: That is what you do in Java

python/hsfs/core/query.py
```diff
@@ -68,13 +85,26 @@ def join(self, sub_query, on=[], left_on=[], right_on=[], join_type="inner"):
         )
         return self
 
+    def as_of(self, wallclocktime):
+        for join in self._joins:
+            join.query.left_featuregroup_endtime = wallclocktime
+        self.left_featuregroup_endtime = wallclocktime
+        return self
+
+    def pull_changes(self, wallclock_starttime, wallclock_endtime):
```
SirOibaf, line 94: snake case

python/hsfs/core/query.py
```diff
@@ -85,3 +115,26 @@ def to_string(self, storage="offline"):
...
+    @left_featuregroup_starttime.setter
+    def left_featuregroup_starttime(self, left_featuregroup_starttime):
+        self._left_featuregroup_starttime = left_featuregroup_starttime
+
+    @left_featuregroup_endtime.setter
+    def left_featuregroup_endtime(self, left_featuregroup_starttime):
+        self._left_featuregroup_endtime = left_featuregroup_starttime
+
+
+def _register_hudi_tables(hudi_feature_groups, feature_store_id, feature_store_name):
```
SirOibaf, line 136: Make this a class method like the rest.

python/hsfs/core/hudi_engine.py
```diff
@@ -0,0 +1,195 @@
...
+
+        return self._feature_group_api.commit(self._feature_group, fg_commit)
+
+    def register_temporary_table(self, alias, start_timestamp, end_timestamp):
+        self._spark_session.conf.set("spark.sql.hive.convertMetastoreParquet", "false")
+
+        hudi_options = self._setup_hudi_read_opts(start_timestamp, end_timestamp)
+        self._spark_session.read.format(self.HUDI_SPARK_FORMAT).options(
+            **hudi_options
+        ).load(self._base_path).registerTempTable(alias)
```
SirOibaf, line 102: `registerTempTable` is deprecated, you should use `createOrReplaceTempView` instead. Which also takes care of overwriting the alias in case you run multiple queries in the same Spark session.

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
```diff
@@ -0,0 +1,214 @@
...
+
+  public void registerTemporaryTable(SparkSession sparkSession, FeatureGroup featureGroup, String alias,
+                                     Long startTimestamp, Long  endTimestamp) {
+    sparkSession.conf().set("spark.sql.hive.convertMetastoreParquet", "false");
+
+    Map<String, String> hudiArgs = setupHudiReadOpts(startTimestamp, endTimestamp);
+    sparkSession.read()
+        .format(HUDI_SPARK_FORMAT)
+        .options(hudiArgs)
+        .load(featureGroup.getLocation()).registerTempTable(alias);
```
SirOibaf, line 137: `registerTempTable` is deprecated, you should use `createOrReplaceTempView` instead, which takes care of replacing the alias if you run multiple queries in the same Spark session.

python/hsfs/feature_group.py
```diff
@@ -48,6 +49,7 @@ def __init__(
         feat_hist_enabled=None,
         statistic_columns=None,
         online_enabled=False,
+        time_travel_format=HUDI,
```
SirOibaf, line 52: Does this show up in the documentation?

SirOibaf (changes requested):
Python reviewed.

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
```diff
@@ -0,0 +1,214 @@
...
+
+    FeatureGroupCommit fgCommit = getLastCommitMetadata(sparkSession, featureGroup.getLocation());
+
+    featureGroupApi.featureGroupCommit(featureGroup, fgCommit);
+  }
+
+  public FeatureGroupCommit deleteRecord(SparkSession sparkSession, FeatureGroup featureGroup, Dataset<Row> deleteDF)
+      throws IOException, FeatureStoreException {
+
+    Map<String, String> hudiArgs = setupHudiWriteOpts(featureGroup, HudiOperationType.UPSERT);
```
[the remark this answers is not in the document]
davitbzh: done

python/hsfs/core/hudi_engine.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def _get_cert_pw(self):
+        pwd_path = Path("material_passwd")
+        if not pwd_path.exists():
+            username = os.environ["HADOOP_USER_NAME"]
+            material_directory = Path(os.environ["MATERIAL_DIRECTORY"])
+            pwd_path = material_directory.joinpath(username + "__cert.key")
+
+        with pwd_path.open() as f:
+            return f.read()
```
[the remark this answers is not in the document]
davitbzh: done

python/hsfs/core/hudi_engine.py
```diff
@@ -0,0 +1,195 @@
...
+    ):
+        self._feature_group = feature_group
+        self._spark_context = spark_context
+        self._spark_session = spark_session
+        self._feature_store_id = feature_store_id
+        self._feature_store_name = feature_store_name
+        self._base_path = self._feature_group.location
+        self._table_name = feature_group.name + "_" + str(feature_group.version)
+
+        # TODO (Davit): Need to fix ComplexKeyGenerator bug
```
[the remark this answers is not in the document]
davitbzh: I needed to fix CustomKeyGenerator bug in hudi itself. It works now

python/hsfs/core/hudi_engine.py
```diff
@@ -0,0 +1,195 @@
...
+        self._primary_key = feature_group.primary_key[0]
+        self._partition_key = feature_group.partition_key[0]
+        self._precombine_key = feature_group.partition_key[0]
```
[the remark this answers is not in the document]
davitbzh: no I am using 1st partition key as recombine key. we need to discuss where to store precombine in backend

python/hsfs/core/query.py
davitbzh: done

python/hsfs/core/hudi_engine.py
davitbzh: done

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
davitbzh: done

python/hsfs/feature_group.py
davitbzh: yes

python/hsfs/core/query.py
davitbzh: done

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
davitbzh: done

python/hsfs/core/hudi_engine.py
davitbzh: done

python/hsfs/core/hudi_engine.py
davitbzh: done

python/hsfs/core/query.py
davitbzh: done

python/hsfs/core/query.py
davitbzh: done

python/hsfs/core/query.py
davitbzh: done

python/hsfs/core/hudi_engine.py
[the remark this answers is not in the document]
davitbzh: Ok form now I will use 1st partition key as recombine key. But I think we need new table to handle metadata like this once we add new frameworks.

python/hsfs/core/hudi_engine.py
davitbzh: done

SirOibaf (approved):

Issue #21, opened 2020-03-10
API for getting the label of a training dataset


moritzmeister:
Jim asked for a way to specify the label in a training dataset. And subsequently getting the name/type of it with an API call.

Just putting this here to not forget

PR #126, opened 2020-10-29
[HOPSWORKS-2091] Label/Prediction feature metadata for training datasets

moritzmeister:


davitbzh:
Looks good to me. I think I need to update TFDataEngine according this PR. There label is required to provide. If its already set here then we should not ask user to set label there as well

davitbzh (commented):
Looks good to me. I think I need to update TFDataEngine according this PR. There label is required to provide. If its already set here then we should not ask user to set label there as well. Will do it after this is merged.

moritzmeister:
closes #21

SirOibaf (approved):

PR #132, opened 2020-11-05
[HOPSWORKS-1374] Append: get commit details in correct order and make upsert default op for fg.insert

davitbzh:


README.md
```diff
@@ -66,7 +66,12 @@ fg.save(dataframe)
 
 Upsert new data in to the feature group with `time_travel_format="HUDI"`".
 ```python
-fg.insert(upsert_df, operation="upsert")
+fg.insert(upsert_df, operation)
```
SirOibaf, line 69: `operation` doesn't make much sense here by itself.

docs/index.md
```diff
@@ -66,7 +66,7 @@ fg.save(dataframe)
 
 Upsert new data in to the feature group with `time_travel_format="HUDI"`".
 ```python
-fg.insert(upsert_df, operation="upsert")
+fg.insert(upsert_df, operation)
```
SirOibaf, line 69: Same as above.

java/src/main/java/com/logicalclocks/hsfs/FeatureGroup.java
```diff
@@ -240,29 +240,52 @@ public void save(Dataset<Row> featureData, Map<String, String> writeOptions)
   }
 
   public void insert(Dataset<Row> featureData) throws IOException, FeatureStoreException {
-    insert(featureData, null, false);
+    if (this.timeTravelFormat == TimeTravelFormat.HUDI)  {
+      insert(featureData, null, false, HudiOperationType.UPSERT, null);
+    } else {
+      insert(featureData, null, false);
+    }
```
SirOibaf, line 247: Don't organize the code like this. There is no point of repeating the same if for every  `insert` overwrite method. 

Just do it once in this `insert` - https://github.com/logicalclocks/feature-store-api/blob/7c704e06ac1ede40a01895d6b502f449fd028861/java/src/main/java/com/logicalclocks/hsfs/FeatureGroup.java#L292 

If the operation is `null` and the type of the feature group is `Hudi` then set the operation to upsert.

java/src/main/java/com/logicalclocks/hsfs/FeatureGroup.java
```diff
@@ -333,10 +356,25 @@ public void commitDeleteRecord(Dataset<Row> featureData, Map<String, String> wri
...
   }
 
+  /**
+   * Return commit details.
+   * @throws FeatureStoreException
+   * @throws IOException
+   */
+  public Map<String, Map<String,String>>  commitDetails() throws IOException, FeatureStoreException {
+    return featureGroupEngine.commitDetails(this, null);
+  }
```
SirOibaf, line 376: Invert the order of these 2 functions so that they appear in the documentation from generic to specific

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
```diff
@@ -168,7 +170,7 @@ private FeatureGroupCommit getLastCommitMetadata(SparkSession sparkSession, Stri
     }
 
     // table name
-    String tableName = utils.getTableName(featureGroup);
+    String tableName = featureGroup.getName() + "_" + featureGroup.getVersion();
```
SirOibaf, line 173: Use this method instead: https://github.com/logicalclocks/feature-store-api/blob/7c704e06ac1ede40a01895d6b502f449fd028861/java/src/main/java/com/logicalclocks/hsfs/engine/Utils.java#L112

java/src/main/java/com/logicalclocks/hsfs/engine/FeatureGroupEngine.java
```diff
@@ -185,9 +186,20 @@ public void updateStatisticsConfig(FeatureGroup featureGroup) throws FeatureStor
     featureGroup.setHistograms(apiFG.getHistograms());
   }
 
-  public FeatureGroupCommit[] commitDetails(FeatureGroup featureGroup, Integer limit)
+  public Map<String, Map<String,String>> commitDetails(FeatureGroup featureGroup, Integer limit)
       throws IOException, FeatureStoreException {
-    return featureGroupApi.commitDetails(featureGroup, limit);
+    List<FeatureGroupCommit> featureGroupCommits = featureGroupApi.commitDetails(featureGroup, limit);
```
SirOibaf, line 191: If there are no commits, this is going to be null and the next part of the code will trigger a `NullPointerException`. Check for it.

python/hsfs/feature_group_commit.py
```diff
@@ -38,6 +42,8 @@ def __init__(
     @classmethod
     def from_response_json(cls, json_dict):
         json_decamelized = humps.decamelize(json_dict)
+        if len(json_decamelized["items"]) >= 1:
```
SirOibaf, line 45: there should be a count field returned by the API that lets you know how many items there are in the list. Check if we set the count field in the `CommitDTO` on the Hopsworks side.

SirOibaf (changes requested):

SirOibaf (approved):

PR #133, opened 2020-11-06
[HOPSWORKS-2122] Backend not handling null TimeTravelFormat sent by client 

moritzmeister:


davitbzh (changes requested):
In the beginning idea was to have `HUDI` as default format. However, I get the point that it's the good idea to keep `None` for now until it becomes stable. Can you please add the same with Java client `TimeTravelFormat.NONE` ? So then it's the same in both.

moritzmeister:
@davitbzh this is not the actual default, the default gets set in the `create_feature_group` which is the only way to instantiate a feature group that is not coming from the backend. The default you were setting on the constructor was never going to be used. we should try to keep the defaults in the place where the object actually gets created. so in fact this change doesn't change behaviour. I just removed it to avoid confusion, as it never takes effect anyway.

davitbzh (approved):

Issue #134, opened 2020-11-09
Use Hudi commit ID as statistics timestamp


SirOibaf:
Currently statistics generate their own timestamp. If the feature group is Hudi based, the commit id and the statistics timestamp should be the same.

PR #135, opened 2020-11-09
[HOPSWORKS-2139]  Use Hudi commit ID as statistics timestamp

davitbzh:


moritzmeister:
closes #134

java/src/main/java/com/logicalclocks/hsfs/engine/FeatureGroupEngine.java
```diff
@@ -187,27 +188,46 @@ private void saveOnlineDataframe(FeatureGroup featureGroup, Dataset<Row> dataset
...
+    List<FeatureGroupCommit> featureGroupCommits =
+        featureGroupApi.getCommitDetails(featureGroup, wallclockTimestamp, limit);
     if (featureGroupCommits == null) {
       throw new FeatureStoreException("There are no commit details available for this Feature group");
     }
-    Map<String, Map<String, String>> commitDetails = new HashMap<String, Map<String, String>>();
+    Map<Long, Map<String, String>> commitDetails = new HashMap<>();
     for (FeatureGroupCommit featureGroupCommit : featureGroupCommits) {
-      commitDetails.put(featureGroupCommit.getCommitID().toString(), new HashMap<String, String>() {{
+      commitDetails.put(featureGroupCommit.getCommitID(), new HashMap<String, String>() {{
```
moritzmeister, line 203: why are the values stored as strings, aren't they all integers?

java/src/main/java/com/logicalclocks/hsfs/engine/Utils.java
```diff
@@ -132,4 +136,35 @@ public String getHiveMetastoreConnector(FeatureGroup featureGroup) throws IOExce
     return connStr + "sslTrustStore=t_certificate;trustStorePassword=" + pw
         + ";sslKeyStore=k_certificate;keyStorePassword=" + pw;
   }
+
+  public Long getTimeStampFromDateString(String inputDate) throws FeatureStoreException, ParseException {
+
+    HashMap<Pattern, String> dateFormatPatterns = new HashMap<Pattern, String>() {{
```
moritzmeister, line 142: I am not completely happy about this, looks hard to maintain. This is something DateTimeFormatter should be able to do out of the box https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html

python/hsfs/core/feature_group_api.py
```diff
@@ -200,13 +200,14 @@ def commit(self, feature_group_instance, feature_group_commit_instance):
             ),
         )
 
-    def commit_details(self, feature_group_instance, limit):
+    def commit_details(self, feature_group_instance, wallclock_timestamp, limit):
```
moritzmeister, line 203: In the java part you renamed this function to `get_commit_details`

python/hsfs/feature_group.py
```diff
@@ -903,6 +906,57 @@ def get_validations(self, validation_time=None, commit_time=None):
             self, validation_time, commit_time
         )
 
+    def compute_statistics(self, wallclock_time: Optional[str] = None):
```
moritzmeister, line 909: I think also here this should be in `FeatureGroupBase`

python/hsfs/feature_group.py
```diff
@@ -903,6 +906,57 @@ def get_validations(self, validation_time=None, commit_time=None):
...
+        # Arguments
+            wallclock_time: Date string in the format of "YYYYMMDD" or "YYYYMMDDhhmmss".
+                Only valid if feature group is time travel enabled. If specified will recompute statistics on
+                feature group as of specific point in time. If not specified then will compute statistics
+                as of most recent time of this fg. Defaults to `None`.
+        # Returns
+            `Statistics`. The statistics metadata object.
+        # Raises
+            `RestAPIError`. Unable to persist the statistics.
+        """
```
moritzmeister, line 923:
```suggestion
        """Recompute the statistics for the feature group and save them to the
        feature store.

        Statistics are only computed for data in the offline storage of the feature
        group.

        # Arguments
            wallclock_time: Date string in the format of "YYYYMMDD" or "YYYYMMDDhhmmss".
                Only valid if feature group is time travel enabled. If specified will recompute statistics on
                feature group as of specific point in time. If not specified then will compute statistics
                as of most recent time of this feature group. Defaults to `None`.

        # Returns
            `Statistics`. The statistics metadata object.

        # Raises
            `RestAPIError`. Unable to persist the statistics.
        """
```

moritzmeister (changes requested):
I think you need to move some stuff to the Base classes, the rest are minor questions.

java/src/main/java/com/logicalclocks/hsfs/FeatureGroup.java
```diff
@@ -418,4 +461,30 @@ public FeatureGroupValidation getValidation(Long time, DataValidationEngine.Vali
...
+
+  /**
+   * Recompute the statistics for the feature group and save them to the feature store.
+   *
+   * @param wallclockTime number of commits to return.
+   * @return statistics object of computed statistics
+   * @throws FeatureStoreException
+   * @throws IOException
+   */
+  public Statistics computeStatistics(String wallclockTime) throws FeatureStoreException, IOException, ParseException {
```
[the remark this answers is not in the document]
davitbzh: If user provides `wallclockTime` then it has to be time travel enabled feature group. So client has to check if its timetravel enabled fg and retrieve commit id. `FeatureGroupBase` doesn't have `timeTravelFormat` argument

java/src/main/java/com/logicalclocks/hsfs/FeatureGroup.java
```diff
@@ -418,4 +461,30 @@ public FeatureGroupValidation getValidation(Long time, DataValidationEngine.Vali
...
+    if (this.timeTravelFormat == TimeTravelFormat.NONE) {
+      throw new FeatureStoreException("commitDetails function is only valid for "
```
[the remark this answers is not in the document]
davitbzh: yes it does, there is `computeStatistics()` in base class. But this is for only time travel enabled feature groups

java/src/main/java/com/logicalclocks/hsfs/engine/FeatureGroupEngine.java
davitbzh: We return `commitedOn` as human readable date string. Then to avoid returning as `Object` I convert other values as `String` also

python/hsfs/feature_group.py
davitbzh: I do this because this method is only for time-travel enabled feature groups. I wrote longer explanation above

java/src/main/java/com/logicalclocks/hsfs/FeatureGroup.java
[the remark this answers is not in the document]
moritzmeister: makes sense, I didn't see that :+1:

moritzmeister (approved):

PR #171, opened 2020-11-29
HSFS is defaulting to PyHive instead of Spark on EMR

### LINKED ISSUE #170: HSFS is defaulting to PyHive instead of Spark on EMR
As EMR does not have the /dbfs folder, hsfs is defaulting to use PyHive even though spark is available. I could work around this issue by creating the folder manually.

SirOibaf:
Also remove downloading of the certificates in the connect method for
PySpark clients. Certificates should already be present when the
application is started.

Closes #170


moritzmeister (approved):
If you tested it :+1:

PR #173, opened 2020-12-01
[HOPSWORKS-CLOUD-1205] Add permission simulator to restrictive permis…

maismail:
…sions


steffengr (approved):

berthoug (changes requested):
Is this a mendatory permission? What happens if we don't have it?
I think if it is not mandatory we should not have it here as this should be only the minimum permission requirement.

maismail:
> Is this a mendatory permission? What happens if we don't have it?
> I think if it is not mandatory we should not have it here as this should be only the minimum permission requirement.

It is not mandatory, nothing we skip the permission validation from Hopsworks.ai and proceed normally as we do now. You are right, no need to add it here. I'll close the PR

PR #185, opened 2020-12-06
On demand feature group over S3 buckets

SirOibaf:


SirOibaf:
@moritzmeister - Please have a look at this PR - I still have to test computing statistics over a on-demand feature group.

python/hsfs/storage_connector.py
```diff
@@ -192,6 +192,11 @@ def expiration(self):
         """Cluster temporary credential expiration time."""
         return self._expiration
 
+    @property
+    def bucket(self):
+        """Return the bucket for S3 connectors."""
+        return self._bucket
```
moritzmeister, line 198: should we add the same check as for "path" if it's actually a S3 connector? Since I guess for JDBC a bucket doesn't make sense

python/hsfs/engine/spark.py
```diff
@@ -185,13 +201,7 @@ def save_dataframe(
...
 
     def _save_offline_dataframe(
-        self,
-        table_name,
-        feature_group,
-        dataframe,
-        save_mode,
-        operation,
-        write_options,
+        self, table_name, feature_group, dataframe, save_mode, operation, write_options,
```
moritzmeister, line 204: looks odd to me that black changed this back.

moritzmeister (commented):
few comments

python/hsfs/core/storage_connector_api.py
```diff
@@ -23,13 +23,11 @@ class StorageConnectorApi:
     def __init__(self, feature_store_id):
         self._feature_store_id = feature_store_id
 
-    def get(self, name, connector_type):
+    def get(self, name):
```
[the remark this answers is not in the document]
SirOibaf: I'm torn about this change tbh. So the original issue is that in the Hopsworks schema the assumption was that on-demand feature group were backed by a JDBC connector. 
We want to move past this assumption and make that a on-demand feature group could potentially be backed by any storage connector we have (mainly JDBC and buckets/dirs with files Spark can read). This resulted in this massive PR: https://github.com/logicalclocks/hopsworks-ee/pull/382

Now, with this PR connector names are also unique across the feature store, removing the need to specify the connector type every time you need to do an operation. 
What I was wondering was the following, I was thinking about deprecating the above method and add a second method with just the name, but then Python does not support method overloading. Alternatively we could default `connector_type` to empty string and just ignoring it in the code, thus avoiding breaking the API.

Honestly I'm not sure what's the best way forward, for the moment I just removed it to implement the rest fo the code.

python/hsfs/storage_connector.py
SirOibaf: Yes, we should probably raise errors if the storage connector is used incorrectly. I'll fix that

python/hsfs/engine/spark.py
SirOibaf: I can run black again, but I don't think I changed it manually.

SirOibaf: black changed it.

python/hsfs/storage_connector.py
SirOibaf: Done

python/hsfs/client/hopsworks.py
```diff
@@ -60,6 +60,7 @@ def __init__(self):
         self._session = requests.session()
 
         self._connected = True
+        self._cert_key = None
```
moritzmeister, line 63: Don't forget to remove this again if we merge #201 first.

moritzmeister: @SirOibaf I think you still need to remove this.

moritzmeister (changes requested):
please remove that one line in the hopsworks client

SirOibaf:
> please remove that one line in the hopsworks client

done

moritzmeister (approved):
:+1:

PR #193, opened 2020-12-10
[HOPSWORKS-2177] Refactor storage connectors

SirOibaf:


java/src/main/java/com/logicalclocks/hsfs/FeatureStore.java
```diff
@@ -121,9 +121,8 @@ public OnDemandFeatureGroup getOnDemandFeatureGroup(String name) throws FeatureS
     return SparkEngine.getInstance().sql(query);
   }
 
-  public StorageConnector getStorageConnector(String name, StorageConnectorType type)
-      throws FeatureStoreException, IOException {
-    return storageConnectorApi.getByNameAndType(this, name, type);
+  public StorageConnector getStorageConnector(String name) throws FeatureStoreException, IOException {
+    return storageConnectorApi.getByNameAndType(this, name);
```
moritzmeister, line 125: Can we call it `getByName` to make it less confusing

```suggestion
    return storageConnectorApi.getByName(this, name);
```

java/src/main/java/com/logicalclocks/hsfs/engine/Utils.java
```diff
@@ -116,7 +114,7 @@ public String getFgName(FeatureGroup featureGroup) {
 
   public String getHiveMetastoreConnector(FeatureGroup featureGroup) throws IOException, FeatureStoreException {
     StorageConnector storageConnector = storageConnectorApi.getByNameAndType(featureGroup.getFeatureStore(),
```
moritzmeister, line 116:
```suggestion
    StorageConnector storageConnector = storageConnectorApi.getByName(featureGroup.getFeatureStore(),
```

java/src/main/java/com/logicalclocks/hsfs/metadata/StorageConnectorApi.java
```diff
@@ -36,7 +35,7 @@
 
   private static final Logger LOGGER = LoggerFactory.getLogger(StorageConnectorApi.class);
 
-  public StorageConnector getByNameAndType(FeatureStore featureStore, String name, StorageConnectorType type)
+  public StorageConnector getByNameAndType(FeatureStore featureStore, String name)
```
moritzmeister, line 38:
```suggestion
  public StorageConnector getByName(FeatureStore featureStore, String name)
```

python/hsfs/storage_connector.py
```diff
@@ -110,77 +110,167 @@ def connector_type(self):
     @property
     def access_key(self):
         """Access key."""
-        return self._access_key
+        if self._storage_connector_type == "S3":
```
moritzmeister, line 113: The user is not able to set these properties in any way right? they are always coming in capital letters from the backend?
Just want to make sure we don't need to lower/upper case these things.

python/hsfs/storage_connector.py
```diff
@@ -190,17 +280,59 @@ def iam_role(self):
     @property
     def expiration(self):
         """Cluster temporary credential expiration time."""
-        return self._expiration
+        if (
+            self._storage_connector_type == "S3"
+            or self._storage_connector_type == "REDSHIFT"
+        ):
```
moritzmeister, line 286: simplify

```suggestion
        if self._storage_connector_type in ["S3", "REDSHIFT"]:
```

moritzmeister (changes requested):
Some naming changes

python/hsfs/storage_connector.py
SirOibaf: Yes, but it's easy enough to add an upper case. Let me do it .

SirOibaf: Done

moritzmeister (approved):
:+1:  just make sure you have replaced the `getByNameAndType` everywhere with `getByName`, I only marked the occurences that I saw here in the code that the PR changes.

PR #194, opened 2020-12-10
[HOPSWORKS-2134] Extend Query constructor with filter capability

moritzmeister:


java/src/main/java/com/logicalclocks/hsfs/Feature.java
```diff
@@ -102,4 +113,52 @@ public Feature(String name, String type, String onlineType, Boolean primary, Boo
     this.partition = partition;
     this.defaultValue = defaultValue;
   }
+
+  public Filter lt(String value) {
```
SirOibaf, line 117: Why not have a generic "Object" for each condition and then call "toString"? In any case the value is passed to the backend as String. 
As for instance, I'm not sure what happens if you pass an integer or a double in these methods.

SirOibaf (changes requested):

java/src/main/java/com/logicalclocks/hsfs/Feature.java
moritzmeister: yes you are right, I wanted to ask you about that, because I ran into a problem when passing a Float, scala was complaining. will fix it.

SirOibaf (approved):

PR #196, opened 2020-12-11
Allow to set engine in connection

moritzmeister:
SageMaker also has PySpark installed in all it's kernels, nevertheless if it is connected to an EMR cluster.


moritzmeister:
Tested it on Databricks and Sagemaker. Since sagemaker now also has pyspark in it's kernels, we also need to be able to set the engine there.

SirOibaf (approved):

docs/integrations/sagemaker.md
```diff
@@ -168,7 +168,7 @@ conn = hsfs.connection(
     443,                                # Port to reach your Hopsworks instance, defaults to 443
     'my_project',                       # Name of your Hopsworks Feature Store project
     secrets_store='secretsmanager',     # Either parameterstore or secretsmanager
-    hostname_verification=True)         # Disable for self-signed certificates
+    hostname_verification=True          # Disable for self-signed certificates
```
SirOibaf, line 171: Can you update the documentation and add the `engine` parameter here?

SirOibaf (changes requested):

docs/integrations/sagemaker.md
moritzmeister: added a note and the parameter for the relevant integrations

SirOibaf (approved):

PR #198, opened 2020-12-17
[HOPSWORKS-2188] Remove partition key requirements for HUDI feature groups and allow users to set precombine keys

davitbzh:
also includes https://logicalclocks.atlassian.net/browse/HOPSWORKS-2196

### DIFF
--- a/java/src/main/java/com/logicalclocks/hsfs/FeatureGroup.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/FeatureGroup.java
@@ -85,6 +85,10 @@ public class FeatureGroup extends FeatureGroupBase {
   // These are only used in the client. In the server they are aggregated in the `features` field
   private List<String> partitionKeys;
 
+  @JsonIgnore
+  // This is only used in the client. In the server they are aggregated in the `features` field
+  private String hudiPrecombineKey;
+
   private FeatureGroupEngine featureGroupEngine = new FeatureGroupEngine();
   private StatisticsEngine statisticsEngine = new StatisticsEngine(EntityEndpointType.FEATURE_GROUP);
 
@@ -92,15 +96,17 @@ public class FeatureGroup extends FeatureGroupBase {
 
   @Builder
   public FeatureGroup(FeatureStore featureStore, @NonNull String name, Integer version, String description,
-                      List<String> primaryKeys, List<String> partitionKeys, boolean onlineEnabled,
-                      TimeTravelFormat timeTravelFormat, List<Feature> features, Boolean statisticsEnabled,
-                      Boolean histograms, Boolean correlations, List<String> statisticColumns) {
+                      List<String> primaryKeys, List<String> partitionKeys, String hudiPrecombineKey,
+                      boolean onlineEnabled, TimeTravelFormat timeTravelFormat, List<Feature> features,
+                      Boolean statisticsEnabled, Boolean histograms, Boolean correlations,
+                      List<String> statisticColumns) {
     this.featureStore = featureStore;
     this.name = name;
     this.version = version;
     this.description = description;
     this.primaryKeys = primaryKeys;
     this.partitionKeys = partitionKeys;
+    this.hudiPrecombineKey = timeTravelFormat == TimeTravelFormat.HUDI ? hudiPrecombineKey : null;
     this.onlineEnabled = onlineEnabled;
     this.timeTravelFormat = timeTravelFormat != null ? timeTravelFormat : TimeTravelFormat.HUDI;
     this.features = features;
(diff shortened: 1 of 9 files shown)


java/src/main/java/com/logicalclocks/hsfs/FeatureGroup.java
```diff
@@ -226,6 +233,11 @@ public void insert(Dataset<Row> featureData, boolean overwrite, Map<String, Stri
     insert(featureData, null, overwrite, null, writeOptions);
   }
 
+  public void insert(Dataset<Row> featureData,  Map<String, String> writeOptions)
```
SirOibaf, line 236: Style change: Can you keep the order consistent and move this function up? Functions should be ordered from the "most generic" to the "most specific".

java/src/main/java/com/logicalclocks/hsfs/engine/FeatureGroupEngine.java
```diff
@@ -102,6 +113,19 @@ public void saveFeatureGroup(FeatureGroup featureGroup, Dataset<Row> dataset, Li
...
+      String hudiPrecombineFeatureName = apiFG.getFeatures().stream()
+          .filter(Feature::getHudiPrecombineKey)
+          .map(Feature::getName)
+          .collect(Collectors.toList()).get(0);
+      featureGroup.getFeatures().forEach(f -> {
+        if (f.getName().equals(hudiPrecombineFeatureName)) {
+          f.setHudiPrecombineKey(true);
+        }
+      });
+    }
```
SirOibaf, line 127: Honestly here I'd just replace the feature list directly. It's more efficient since you already have deserialized it.

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
```diff
@@ -155,24 +160,27 @@ private FeatureGroupCommit getLastCommitMetadata(SparkSession sparkSession, Stri
...
+    } else {
+      hudiArgs.put(HUDI_PARTITION_FIELD, "");
+      hudiArgs.put(HIVE_PARTITION_EXTRACTOR_CLASS_OPT_KEY, HIVE_NON_PARTITION_EXTRACTOR_CLASS_OPT_VAL);
     }
 
-    // table name
-    String tableName = utils.getFgName(featureGroup);
-    hudiArgs.put(HUDI_TABLE_NAME, tableName);
-    hudiArgs.put(HIVE_PARTITION_EXTRACTOR_CLASS_OPT_KEY, DEFAULT_HIVE_PARTITION_EXTRACTOR_CLASS_OPT_VAL);
+    List<Feature> precombineKey = featureGroup.getFeatures().stream().filter(Feature::getHudiPrecombineKey)
```
SirOibaf, line 180: Why is this a list now? In the rest of the code precombine is a single column, Just use getFirst on the stream, no point in building a list when you know you'll have a single element

python/hsfs/feature.py
```diff
@@ -45,6 +46,7 @@ def __init__(
         self._description = description
         self._primary = primary or False
         self._partition = partition or False
+        self._hudi_precombine_key = hudi_precombine_key or False
```
SirOibaf, line 49: You don't need to the `or False` here.

python/hsfs/feature_store.py
```diff
@@ -229,6 +229,7 @@ def create_feature_group(
         time_travel_format: Optional[str] = "HUDI",
         partition_key: Optional[List[str]] = [],
         primary_key: Optional[List[str]] = [],
+        hudi_precombine_key: Optional[List[str]] = None,
```
SirOibaf, line 232: Why is this a list here?

python/hsfs/core/feature_group_engine.py
```diff
@@ -38,9 +38,23 @@ def save(self, feature_group, feature_dataframe, write_options):
                 feat.primary = True
             if feat.name in feature_group.partition_key:
                 feat.partition = True
+            if (
+                feature_group.hudi_precombine_key is not None
+                and feat.name in feature_group.hudi_precombine_key
```
SirOibaf, line 43: Why using the `in` here. precombine_key can be at most one string, just compare 2 strings.

python/hsfs/core/feature_group_engine.py
```diff
@@ -38,9 +38,23 @@ def save(self, feature_group, feature_dataframe, write_options):
...
 
+        # if hudi precombine key was not provided and TimeTravelFormat is HUDI, retrieve from backend and set
+        if (
+            feature_group.hudi_precombine_key is None
+            and feature_group.time_travel_format == "HUDI"
+        ):
+            for feat in self.features:
+                if feat.hudi_precombine_key:
+                    feature_group.hudi_precombine_key = feat.name
+
```
SirOibaf, line 57: You shouldn't need this step, the Python feature_group_api automatically updates the `feature_group` object with what comes back from the backend.

python/hsfs/feature_group.py
```diff
@@ -277,11 +276,17 @@ def __init__(
                 feat.name for feat in self._features if feat.partition is True
             ]
 
+            self._hudi_precombine_key = [
+                feat.name for feat in self._features if feat.hudi_precombine_key is True
+            ]
```
SirOibaf, line 281: again. why a list when we always consider a single value?

SirOibaf (changes requested):

SirOibaf (approved):

PR #206, opened 2021-01-03
[HOPSWORKS-1971] Add HSFS python support

SirOibaf:


utils/hsfs_utils.py
```diff
@@ -0,0 +1,109 @@
+import argparse
+import json
+import hsfs
+
+from typing import Dict, Any
+from pydoop import hdfs
+from pyspark.sql import SparkSession
```
moritzmeister, line 7: I believe these imports will lead to problems on environments without pyspark and pydoop installed. Haven't tried yet though.

moritzmeister (changes requested):

utils/hsfs_utils.py
moritzmeister: ah sorry didn't see it's outside the python directory

moritzmeister: ah sorry didn't see the file is outside the python package

python/hsfs/core/ingestion_job_conf.py
```diff
@@ -0,0 +1,59 @@
...
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+
+import json
+from hsfs import util
+
+
+class IngestionJob:
```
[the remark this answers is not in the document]
SirOibaf: done

python/hsfs/core/training_dataset_engine.py
```diff
@@ -29,15 +30,31 @@ def __init__(self, feature_store_id):
...
-    def save(self, training_dataset, feature_dataframe, user_write_options):
+    def save(self, training_dataset, features, user_write_options):
+        if isinstance(features, query.Query):
+            training_dataset._querydto = features
+            training_dataset._features = [
+                training_dataset_feature.TrainingDatasetFeature(
+                    name=label_name, label=True
+                )
+                for label_name in training_dataset.label
+            ]
```
moritzmeister, line 41: This part I don't understand. Don't you have only the label features in the `features` property then? But it should be the entire schema of the training dataset. Also why saving the schema in `_querydto`?

SirOibaf: Yes, only the labels will be sent and they will be validated in the backend.

moritzmeister: hm okay

python/hsfs/feature_group.py
```diff
@@ -562,7 +573,9 @@ def save(
 
         user_version = self._version
         self._feature_group_engine.save(self, feature_dataframe, write_options)
-        if self.statistics_config.enabled:
+        if self.statistics_config.enabled and engine.get_type() == "spark":
```
moritzmeister, line 576: Don't you have the `engine.get_type() == "spark"` also in the `statistics_engine.compute_statistics()`?

SirOibaf: Yes it's there as well. However the main issue here is that, without that flag, the code would trigger two jobs
- the first to insert data in the feature group
- the second to compute the statistics (which is trigger by the `compute_statistics` method).

The second job is not needed as the statistics are already computed by the first job. That's why I put that flag there.

moritzmeister (approved):

PR #213, opened 2021-01-07
compute_statistics

davitbzh:


python/hsfs/feature_group.py
```diff
@@ -640,6 +648,27 @@ def commit_delete_record(
         """
         self._feature_group_engine.commit_delete(self, delete_df, write_options)
 
+    def compute_statistics(self):
```
SirOibaf, line 651: This is wrong. `compute_statistics` should be a member of the `FeatureGroupBase` class so that it can be used also by the on-demand feature groups

SirOibaf (changes requested):

python/hsfs/feature_group.py
```diff
@@ -218,6 +218,27 @@ def get_feature(self, name: str):
                 f"'FeatureGroup' object has no feature called '{name}'."
             )
 
+    def compute_statistics(self):
```
SirOibaf, line 221: Can you please not mix the methods? The rest of the statistics methods start at line 265.

SirOibaf (changes requested):

SirOibaf (approved):

PR #216, opened 2021-01-13
[HOPSWORKS-1978] schematized tags

o-alex:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-1978

### DIFF
--- a/python/hsfs/core/tags_api.py
+++ b/python/hsfs/core/tags_api.py
@@ -54,7 +54,8 @@ class TagsApi:
             "tags",
             name,
         ]
-        _client._send_request("PUT", path_params, data=value)
+        headers = {"content-type": "application/json"}
+        _client._send_request("PUT", path_params, headers=headers, data=value)
 
     def delete(self, metadata_instance, name):
         """Delete a tag from a training dataset or feature group.
--- a/python/hsfs/core/tags_api.py
+++ b/python/hsfs/core/tags_api.py
@@ -14,6 +14,8 @@
 #   limitations under the License.
 #
 
+import json
+
 from hsfs import client, tag
 
 
@@ -55,7 +57,8 @@ class TagsApi:
             name,
         ]
         headers = {"content-type": "application/json"}
-        _client._send_request("PUT", path_params, headers=headers, data=value)
+        json_value = json.dumps(value)
+        _client._send_request("PUT", path_params, headers=headers, data=json_value)
 
     def delete(self, metadata_instance, name):
         """Delete a tag from a training dataset or feature group.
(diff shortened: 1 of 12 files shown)


o-alex:
@SirOibaf can you please change my rights so I can add reviewers.
Thanks

python/hsfs/feature_group.py
```diff
@@ -143,11 +143,11 @@ def filter(self, f: Union[filter.Filter, filter.Logic]):
...
         return self.select_all().filter(f)
 
-    def add_tag(self, name: str, value: str = None):
+    def add_tag(self, name: str, value: str):
         """Attach a name/value tag to a feature group.
 
-        A tag can consist of a name only or a name/value pair. Tag names are
-        unique identifiers.
+        A tag consists of a name/value pair. Tag names are unique identifiers.
+        The value of a tag can be any valid json - primitives, arrays or json objects.
```
SirOibaf, line 150: `json - primitives, arrays or json objects.` this is not true as you are explicitly asking for a str. I think it would be better if we accept both `str` and `Dict` here and then check the type of the value.
If the value is `Dict` we convert it to json, if it's string, we just send it to the backend.

python/hsfs/core/tags_api.py
```diff
@@ -54,8 +54,7 @@ def add(self, metadata_instance, name, value):
             "tags",
             name,
         ]
-        query_params = {"value": value} if value else None
-        _client._send_request("PUT", path_params, query_params=query_params)
+        _client._send_request("PUT", path_params, data=value)
```
SirOibaf, line 57: Did you test this? You are not setting the JSON header. What happens if the value provided by the user is not a valid JSON?

python/hsfs/training_dataset.py
```diff
@@ -286,11 +286,11 @@ def show(self, n: int, split: str = None):
         """
         self.read(split).show(n)
 
-    def add_tag(self, name: str, value: str = None):
+    def add_tag(self, name: str, value: str):
```
SirOibaf, line 289: Same comment for feature groups.

SirOibaf (changes requested):
Beside the comments, you didn't change the Java part.

python/hsfs/core/tags_api.py
o-alex: Added the header. 
The backend deals with malformed tags - malformed json or value that does not match schema.

python/hsfs/feature_group.py
o-alex: removed the str constrain, allowing any object that can have `json.dumps` applied to it correctly
adding a ruby test in hopsworks for this...

python/hsfs/training_dataset.py
o-alex: removed the `str` constrain, allowing any object that can have `json.dumps` applied to it correctly
adding a ruby test in hopsworks for this...

SirOibaf (approved):

PR #221, opened 2021-01-18
[HOPSWORKS-2197] Add ADLS connector support

SirOibaf:


python/hsfs/engine/spark.py
```diff
@@ -419,7 +417,15 @@ def training_dataset_schema_match(self, dataframe, schema):
...
+    def _setup_storage_connector(self, storage_connector, path=None):
+        if storage_connector.connector_type == StorageConnector.S3:
+            return self._setup_s3_hadoop_conf(storage_connector, path)
+        elif storage_connector.connector_type == StorageConnector.ADLS:
+            return self._setup_adls_hadoop_conf(storage_connector, path)
+        else:
+            return path
+
+    def _setup_s3_hadoop_conf(self, storage_connector, path):
         if storage_connector.access_key:
```
moritzmeister, line 429: these if statements look different to the java part. There you also check based on the session token.

python/hsfs/storage_connector.py
```diff
@@ -363,6 +454,8 @@ def spark_options(self):
             if self._table_name is not None:
                 props["dbtable"] = self._table_name
             return props
+        elif self._storage_connector_type.upper() == self.ADLS:
```
moritzmeister, line 457: can you replace the ifs above also with the class variables, e.g. `== self.REDSHIFT` for consistency

python/hsfs/engine/spark.py
SirOibaf: I just moved the code tbh. But I see your point to keep it equals. I changed the Java code to match the Python code. 
Doing validation at that validation of the connector values at that point in the pipeline doesn't make much sense and some of the configurations (access key and secret key) were repeated in the function.

PR #232, opened 2021-01-29
[HOPSWORKS-2341] allow hsfs to get all metadata from feature store within experiment function

davitbzh:


python/hsfs/connection.py
```diff
@@ -184,6 +184,8 @@ def connect(self):
                 self._engine is None and not importlib.util.find_spec("pyspark")
             ):
                 self._engine = "hive"
+            elif self._engine is not None and self._engine.lower() == "hops_experiment":
```
moritzmeister, line 187: I wouldn't necessarily call it "hops-experiment", but I also have no strong opinion what else it could be. so fine with me.

python/hsfs/engine/__init__.py
```diff
@@ -39,6 +39,8 @@ def init(engine_type):
                 )
             _engine_type = "hive"
             _engine = hive.Engine()
+        elif engine_type == "hops_experiment":
+            pass
```
moritzmeister, line 43: how does the error look like if you try to call something that requires an engine? we could add a exception that with some info in the `get_instance()` if `_engine == "hops-experiment"`. just to avoid cryptic error messages.

moritzmeister (approved):
Some documentation would be good, if it has to go in quickly, it's fine!

python/hsfs/connection.py
davitbzh: may be `tensorflow` or `tf_data`?

SirOibaf:
Actually this PR doesn't work. The last append was not tested, nowhere we assign `_engine = "training"`, so in the `get_instance()` the inner if statement is never triggered. cc / @davitbzh

PR #238, opened 2021-02-05
[HOPSWORKS-2458] Online-fs Kafka ingestion

moritzmeister:


python/hsfs/feature_group.py
```diff
@@ -380,13 +381,14 @@ def __init__(
         created=None,
         creator=None,
         id=None,
-        features=None,
+        features=[],
```
SirOibaf, line 384: Revert to none to avoid static lists.

SirOibaf (changes requested):

python/hsfs/feature_group.py
moritzmeister: fixed

SirOibaf (approved):

PR #251, opened 2021-02-19
Add auto-lower casing for feature names

### LINKED ISSUE #178: Add auto-rename for feature names

moritzmeister:
closes #178


java/src/main/java/com/logicalclocks/hsfs/engine/Utils.java
```diff
@@ -65,12 +68,18 @@
...
       // TODO(Fabio): unit test this one for complext types
-      features.add(new TrainingDatasetFeature(structField.name(), structField.dataType().catalogString(), index++));
+      features.add(new TrainingDatasetFeature(
+          structField.name().toLowerCase(), structField.dataType().catalogString(), index++));
     }
 
     return features;
   }
 
+  public Dataset<Row> sanetizeFeatureNames(Dataset<Row> dataset) {
```
SirOibaf, line 78: typo

java/src/main/java/com/logicalclocks/hsfs/engine/Utils.java
```diff
@@ -65,12 +68,18 @@
...
+    return dataset.select(Arrays.asList(dataset.columns()).stream().map(f -> col(f).alias(f.toLowerCase())).toArray(
+        Column[]::new));
```
SirOibaf, line 80: You can actually just leave them strings instead of converting them to column, select should work just fine

SirOibaf (changes requested):
What about select statements in the queries. Should we assume that the user is going to write lower case, or should we lower case them as well?

java/src/main/java/com/logicalclocks/hsfs/engine/Utils.java
moritzmeister: I know it looks stupid, but because of method overloading there is no select which takes just a single `String[]`. Instead the signature is `public Dataset<Row> select(String col, String... cols)` so I would have had to split the array first. I thought this is cleaner.

java/src/main/java/com/logicalclocks/hsfs/engine/Utils.java
moritzmeister: fixed

java/src/main/java/com/logicalclocks/hsfs/FeatureGroup.java
```diff
@@ -119,9 +120,10 @@ public FeatureGroup(FeatureStore featureStore, @NonNull String name, Integer ver
...
-    this.primaryKeys = primaryKeys;
-    this.partitionKeys = partitionKeys;
-    this.hudiPrecombineKey = timeTravelFormat == TimeTravelFormat.HUDI ? hudiPrecombineKey : null;
+    this.primaryKeys = primaryKeys.stream().map(String::toLowerCase).collect(Collectors.toList());
+    this.partitionKeys = partitionKeys.stream().map(String::toLowerCase).collect(Collectors.toList());
+    this.hudiPrecombineKey = timeTravelFormat == TimeTravelFormat.HUDI && hudiPrecombineKey != null
+        ? hudiPrecombineKey.toLowerCase() : null;
     this.onlineEnabled = onlineEnabled;
     this.timeTravelFormat = timeTravelFormat != null ? timeTravelFormat : TimeTravelFormat.HUDI;
     this.features = features;
```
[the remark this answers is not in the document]
moritzmeister: lower casing in the instantiation of feature objects now

moritzmeister:
> What about select statements in the queries. Should we assume that the user is going to write lower case, or should we lower case them as well?

@SirOibaf this is fixed by lower casing all names in the constructors of the `Feature` Objects :)

SirOibaf (approved):

PR #258, opened 2021-02-24
Update Contributing Docs for multi-version docs

moritzmeister:


CONTRIBUTING.md
```diff
@@ -87,14 +87,73 @@ We use `mkdocs` to build the documentation and a plugin called `keras-autodoc` t
     python auto_doc.py
     ```
 
+##### Option 1: Build only current version of docs
+
 4. Either build the docs, or serve them dynamically:
 
+    Note: Links and pictures might not resolve properly later on when checking with this build.
```
berthoug, line 94: I would maybe put the reason why they would not: `as they are built to work with the versioning enabled` or something like that.

CONTRIBUTING.md
```diff
@@ -87,14 +87,73 @@ We use `mkdocs` to build the documentation and a plugin called `keras-autodoc` t
...
+    ```bash
+    python auto_doc.py
+    ```
+
+    Build docs with a version and alias
+    ```bash
+    mike deploy [version] [alias] --update-alias
+
+    # for example, if you branch is based on master and becomes new SNAPSHOT version:
+    mike deploy 2.2.0-SNAPSHOT dev --update-alias
```
berthoug, line 123: I have no clue what the version and alias are related to. The comment make it even stranger as you speak of master and `SNAPSHOT` but then have `2.2.0-SNAPSHOT` and `dev`. My brain then says that `2.2.0-SNAPSHOT` is for the `SNAPSHOT` part but this does not make sense as the order does not match the first part `based on master and becomes new SNAPSHOT`
I'm not sure it is clear as it is not clear for myself.
And if `2.2.0-SNAPSHOT` is for master this will not work once the master version get updated.

CONTRIBUTING.md
```diff
@@ -87,14 +87,73 @@ We use `mkdocs` to build the documentation and a plugin called `keras-autodoc` t
...
+    Delete and reset your local gh-pages branch:
+    ```bash
+    mike delete --all
+
+    # or delete single version
+    mike delete [version-or-alias]
+    ```
+
+    **Background about `mike`:**
+    `mike` builds the documentation and commits it as a new directory to the gh-pages branch. Each directory corresponds to one version of the documentation. Additionally, `mike` maintains a json in the root of gh-pages with the mappings of versions/aliases for each of the directories available. With aliases you can define extra names like `dev` or `latest`.
```
berthoug, line 154: I would put that at the beginning of the section as I think it helps to understand what we are doing.

berthoug (changes requested):

CONTRIBUTING.md
moritzmeister: added

CONTRIBUTING.md
moritzmeister: okay I added a small section on how we actually map branches to versions on docs.hopsworks.ai. Please check if it make sit clearer.

CONTRIBUTING.md
moritzmeister: moved to the top

berthoug (approved):

PR #274, opened 2021-03-08
[HOPSWORKS-2424] Attach transformation function to training dataset

davitbzh:


python/hsfs/core/training_dataset_engine.py
```diff
@@ -16,10 +16,23 @@
...
 from sqlalchemy import sql
-
-from hsfs import engine, training_dataset_feature, util
+import cloudpickle
+import mock
+
+try:
+    from pydoop import hdfs
+except ImportError:
+    hdfs = mock.Mock()
```
SirOibaf, line 25: Don't use Pydoop to get the file of the transformation functions, use the REST Api instead. We cannot use Pydoop on Databricks

python/hsfs/training_dataset.py
```diff
@@ -325,6 +331,33 @@ def update_statistics_config(self):
...
+        )
+
+    # TODO (davit): write docs
+    def python_transformation(self, builder_function):
+        self._training_dataset_engine.attach_transformation_function(
+            self, builder_function, "python"
+        )
+        self._transformation_function = (
+            self._training_dataset_engine.get_transformation_function(self)
+        )
```
SirOibaf, line 359: I think these methods are just confusing. I think we should have a single method `attach_transformation_function`. `function_type` should have a default value.

Why do we need to differentiate between python and pyspark?

python/hsfs/training_dataset.py
```diff
@@ -62,6 +62,7 @@ def __init__(
         from_query=None,
         querydto=None,
         label=None,
+        transformation_function_dto=None,
```
SirOibaf, line 65: I think I mentioned also in the online feature store PR. Don't name the attributes `DTO` they just look ugly on the client side. 

Also I'm a bit confused what's going on. You have both the `_transformation_function_dto` and `_transformation_function`. I doubt you need both. 

We should allow users to configure the transformation function when calling the `create_training_dataset`.

We should also allow users to specify multiple transformation functions, i.e. passing an array instead of a single value.

python/hsfs/core/transformation_function_engine.py
```diff
@@ -0,0 +1,149 @@
...
+        transformation_fns = []
+        for transformation_fn_instance in transformation_fn_instances:
+            transformation_fn_dict = {
+                "name": transformation_fn_instance.name,
+                "version": transformation_fn_instance.version,
+                "output_type": transformation_fn_instance.output_type,
+                "transformation_function": transformation_fn_instance,
+            }
+            transformation_fns.append(transformation_fn_dict)
+        return transformation_fns
```
SirOibaf, line 69: See my other comment about returning objects

python/hsfs/feature_store.py
```diff
@@ -534,6 +545,93 @@ def create_expectation(
...
+            numpy.float64,
+            TypeVar("float64"),  # noqa: F821
+            TypeVar("double"),  # noqa: F821
+            datetime.datetime,
+            numpy.datetime64,
+            datetime.date,
+            bool,
+            TypeVar("boolean"),  # noqa: F821
+            TypeVar("bool"),  # noqa: F821
+            numpy.bool,
```
SirOibaf, line 582: Ok, but we also need to make sure this validation happens in the backend as well. Otherwise users will be able to create non-valid transformations from the UI

python/hsfs/engine/spark.py
```diff
@@ -350,6 +350,16 @@ def write_training_dataset(
...
         dataset = self.convert_to_default_dataframe(dataset)
+
+        # apply transformation functions
+        for (
+            feature_name,
+            transformation_fn,
+        ) in training_dataset.transformation_functions.items():
+            dataset = dataset.withColumn(
+                feature_name, transformation_fn.transformation_fn(col(feature_name))
+            )
```
SirOibaf, line 361: Just FYI, `withColumn` is going to break if you have lots of transformations.

python/hsfs/transformation_function.py
```diff
@@ -0,0 +1,262 @@
...
+        # add all imports from module
+        # add original source code that will be used during offline transformations
+        self._source_code_content = json.dumps(
+            {
+                "module_imports": "\n".join(module_imports),
+                "transformer_code": transformer_code,
+            }
+        )
+
+    @staticmethod
```
SirOibaf, line 103: You don't need the `@staticmethod` annotation when you call methods with `self.`

SirOibaf (changes requested):

python/hsfs/core/transformation_function_engine.py
```diff
@@ -0,0 +1,149 @@
...
+                for td_feature in join.query._left_feature_group.features:
+                    if feature_name == td_feature.name:
+                        transfrom_feature = td_feature
+                        break
+
+        if transfrom_feature is None:
+            raise Exception(
+                "Provided feature %s doesn't exist in training dataset query"
+                % feature_name
+            )
```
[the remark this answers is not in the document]
davitbzh: The problem is that currently is designed in a way that transformation function is attached to `TrainingDatasetFeature` object during creation of `TrainingDataset` object. If provided feature name doesn't exist in `Query` object then transformation function will not be attached. I need check this from client side. 

regarding `@staticmethod` if I do `self.` then it complains that is a staticmethod as I don't use any `self` inside the method.

python/hsfs/transformation_function.py
davitbzh: Its fine for me but in that case style check shows warning

python/hsfs/transformation_function.py
```diff
@@ -0,0 +1,262 @@
...
+                        n.func.attr if isinstance(n.func, ast.Attribute) else n.func.id
+                    )
+                else:
+                    name = n.attr if isinstance(n, ast.Attribute) else n.id
+                decorators[node.name].append(name)
+
+        node_iter = ast.NodeVisitor()
+        node_iter.visit_FunctionDef = _visited_import_mode
+        node_iter.visit(ast.parse(inspect.getsource(transformer)))
+        return decorators
```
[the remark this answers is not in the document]
davitbzh: good point. will remove this

python/hsfs/transformation_function.py
```diff
@@ -0,0 +1,262 @@
...
+                + "\n"
+                + transformer_code
+            )
+        else:
+            self._transformer_code = module_imports + "\n" * 2 + transformer_code
+
+        scope = __import__("__main__").__dict__
+        exec(self._transformer_code, scope)
+        self._transformation_fn = eval(self._name, scope)
+        self._transformation_fn._code = self._transformer_code
```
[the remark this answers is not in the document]
davitbzh: No, its the same. I am just rebuilding it from source code

python/hsfs/engine/spark.py
davitbzh: `transformation_fn(col(feature_name)).alias(feature_name) )` this will be more efficient?  I understand that `withColumn` will create new dataframe every time its called since dataframe is immutable. But not sure if `alias` can avoid that

python/hsfs/transformation_function.py
```diff
@@ -0,0 +1,262 @@
...
+                imported_module = node.module
+            else:
+                continue
+
+            for n in node.names:
+                if imported_module:
+                    import_line = "from " + imported_module + " import " + n.name
+                else:
+                    import_line = "import " + n.name + " as " + n.asname
+                imports.append(import_line)
```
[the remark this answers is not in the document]
davitbzh: I think we need to do it. I am not aware of any other way.

python/hsfs/core/transformation_function_engine.py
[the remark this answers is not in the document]
SirOibaf: [Slack conversation]

python/hsfs/transformation_function.py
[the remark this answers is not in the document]
SirOibaf: the method is still here.

python/hsfs/transformation_function.py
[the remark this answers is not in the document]
SirOibaf: ok

python/hsfs/engine/spark.py
SirOibaf: It's ok, it was just a warning. We can fix it if it becomes a problem.

java/src/main/java/com/logicalclocks/hsfs/engine/TrainingDatasetEngine.java
```diff
@@ -294,4 +306,19 @@ private Object deserializeComplexFeature(Map<String, DatumReader<Object>> comple
   public void delete(TrainingDataset trainingDataset) throws FeatureStoreException, IOException {
     trainingDatasetApi.delete(trainingDataset);
   }
+
+  private List<TrainingDatasetFeature> getTransformationFunction(TrainingDataset trainingDataset)
```
SirOibaf, line 310: In python it's called (more appropriately IMHO) - `getTransformationFunctions` - you return a list here, it should be plural.

python/hsfs/core/transformation_function_api.py
```diff
@@ -0,0 +1,117 @@
...
+            _client._project_id,
+            "featurestores",
+            self._feature_store_id,
+            "transformationfunctions",
+            transformation_function_instance.id,
+        ]
+        headers = {"content-type": "application/json"}
+        _client._send_request("DELETE", path_params, headers=headers)
+
+    def get_training_dataset_transformation_fn(self, training_dataset_instance):
```
SirOibaf, line 97: Why do we still have these long methods names?

python/hsfs/core/training_dataset_engine.py
```diff
@@ -152,6 +171,9 @@ def get_serving_vector(self, training_dataset, entry, external):
         # get schemas for complex features once
         complex_features = self.get_complex_feature_schemas(training_dataset)
 
+        # get transformation functions
+        transformation_fns = self._get_transformation_fns(training_dataset)
```
SirOibaf, line 175: This should happen in the init statement, not here. otherwise the for loop to reload the code is run at every request.

python/hsfs/core/transformation_function_engine.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+        transformation_fns = []
+        for transformation_fn_instance in transformation_fn_instances:
+            transformation_fns.append(transformation_fn_instance)
+        return transformation_fns
+
+    def delete(self, transformation_function_instance):
+        self._transformation_function_api.delete(transformation_function_instance)
+
+    def get_training_dataset_transformation_fn(self, training_dataset):
```
SirOibaf, line 56: rename this method.

SirOibaf (changes requested):

SirOibaf:
Please write documentation.

python/hsfs/core/transformation_function_api.py
davitbzh: there is `get_transformation_fn` already but I will rename it as `get_td_transformation_fn`

SirOibaf (approved):

PR #279, opened 2021-03-10
Fix Hudi Cert Key issue External Clients

moritzmeister:
JAVA NOT TESTED YET!

### DIFF
--- a/java/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksExternalClient.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksExternalClient.java
@@ -65,10 +65,6 @@ public class HopsworksExternalClient implements HopsworksHttpClient {
   private static final String PARAM_NAME_SECRET_STORE = "hopsworks/role/";
   private static final String PARAM_NAME_PARAMETER_STORE = "/hopsworks/role/";
 
-  private static final String MATERIAL_PASSWD = "material_passwd";
-  private static final String T_CERTIFICATE = "t_certificate";
-  private static final String K_CERTIFICATE = "k_certificate";
-
   private PoolingHttpClientConnectionManager connectionPool = null;
 
   private HttpHost httpHost = null;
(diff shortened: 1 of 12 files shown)


java/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksExternalClient.java
```diff
@@ -63,13 +65,26 @@
   private static final String PARAM_NAME_SECRET_STORE = "hopsworks/role/";
   private static final String PARAM_NAME_PARAMETER_STORE = "/hopsworks/role/";
 
+  private static final String MATERIAL_PASSWD = "material_passwd";
+  private static final String T_CERTIFICATE = "t_certificate";
+  private static final String K_CERTIFICATE = "k_certificate";
```
SirOibaf, line 70: What are these for?

java/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksHttpClient.java
```diff
@@ -19,10 +19,34 @@
...
+      int content;
+      while ((content = fis.read()) != -1) {
+        sb.append((char) content);
+      }
+      return sb.toString();
+    } catch (IOException ex) {
+      LOGGER.warn("Failed to get cert password.", ex);
+    }
+    return null;
+  }
```
SirOibaf, line 51: Use the commons-io library, there is a FileUtils.readFileToString method which makes reading a String from a file a one-liner.

SirOibaf (changes requested):

java/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksExternalClient.java
moritzmeister: copy paste, removed them

java/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksHttpClient.java
moritzmeister: replaced as you said, dunno where I got this monster from

SirOibaf (approved):

SirOibaf:
Approved

PR #297, opened 2021-03-31
[HOPSWORKS-2471] remove quotes from prepared statement parameter

davitbzh:


SirOibaf:
Why sending the quotes in the first place?

java/src/main/java/com/logicalclocks/hsfs/engine/TrainingDatasetEngine.java
```diff
@@ -220,6 +220,10 @@ public void initPreparedStatement(TrainingDataset trainingDataset)
     // init prepared statement if it has not already
     if (trainingDataset.getPreparedStatements() == null) {
       initPreparedStatement(trainingDataset);
+      //check if primary key map correspond to serving_keys. We check this during init state to reduce latency.
+      if (!trainingDataset.getServingKeys().equals(entry.keySet())) {
+        throw new IllegalArgumentException("Provided primary key map doesn't correspond to serving_keys");
+      }
```
SirOibaf, line 226: This doesn't make sense, as the user can potentially provide a different set of keys at each method call.

python/hsfs/core/training_dataset_engine.py
```diff
@@ -123,6 +123,11 @@ def get_serving_vector(self, training_dataset, entry):
 
         if training_dataset.prepared_statements is None:
             self.init_prepared_statement(training_dataset)
+            # check if primary key map correspond to serving_keys. We check this during init state to reduce latency.
+            if not entry.keys() == training_dataset.serving_keys:
+                raise ValueError(
+                    "Provided primary key map doesn't correspond to serving_keys"
+                )
```
SirOibaf, line 130: Same comment as for Java

SirOibaf (changes requested):

python/hsfs/core/training_dataset_engine.py
davitbzh: set of key names, which are primary key names, has to be always the same, otherwise how application knows which feature vector to retrieve?

java/src/main/java/com/logicalclocks/hsfs/engine/TrainingDatasetEngine.java
davitbzh: commented bellow but I am just checking map key names which has to be always the same

SirOibaf (approved):

PR #313, opened 2021-04-16
Subclass the different storage connector types

### LINKED ISSUE #219: Cleanup storage connector code
Right now we have a single class representing all types of storage connectors. 
As we keep adding new storage connectors, this class is becoming quite a mess. 
We should somehow split it up in smaller classes.

moritzmeister:
closes #219


java/src/main/java/com/logicalclocks/hsfs/engine/Utils.java
```diff
@@ -141,7 +141,8 @@ public String getFgName(FeatureGroup featureGroup) {
   }
 
   public String getHiveMetastoreConnector(FeatureGroup featureGroup) throws IOException, FeatureStoreException {
```
SirOibaf, line 143: This method name is wrong. The connection is not for the metastore, but for the hiveserver

python/hsfs/engine/spark.py
```diff
@@ -102,14 +102,14 @@ def set_job_group(self, group_id, description):
 
     def register_on_demand_temporary_table(self, on_demand_fg, alias, options={}):
```
SirOibaf, line 103: Why not following the same pattern as for Java?

python/hsfs/storage_connector.py
```diff
@@ -14,38 +14,197 @@
...
+        """If the connector refers to a path (e.g. S3) - return the path of the connector"""
+        return "s3://" + self._bucket
+
+    def spark_options(self):
+        """Return prepared options to be passed to Spark, based on the additional
+        arguments.
+        """
+        return {}
+
+    def prepare_spark(self, path: Optional[str] = None):
```
SirOibaf, line 180: where is this call being used?

python/hsfs/storage_connector.py
```diff
@@ -336,300 +294,220 @@ def iam_role(self):
...
-
-            return props
-        else:
-            raise Exception(
-                "Spark options are not supported for connector "
-                + self._storage_connector_type
-            )
+        return self._options
 
     def snowflake_connector_options(self):
```
SirOibaf, line 512: Is this method being used anywhere?

SirOibaf (changes requested):

SirOibaf:
Please port the changes of this PR: https://github.com/logicalclocks/feature-store-api/pull/344/

java/src/main/java/com/logicalclocks/hsfs/StorageConnector.java
```diff
@@ -36,247 +41,285 @@
...
+    @Getter @Setter
+    private String containerName;
+
+    @Getter @Setter
+    private List<Option> sparkOptions;
+
+    public Map<String, String> sparkOptions() {
+      Map<String, String> options = new HashMap<>();
+      sparkOptions.stream().forEach(option -> options.put(option.getName(), option.getValue()));
+      return options;
```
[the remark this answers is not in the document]
moritzmeister: Made abstract and added the exception instead to the connectors that don't implement the method.

java/src/main/java/com/logicalclocks/hsfs/engine/Utils.java
moritzmeister: fix by applying #344

java/src/main/java/com/logicalclocks/hsfs/StorageConnector.java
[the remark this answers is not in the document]
moritzmeister: actually, I noticed. that before there was a default for the rest of the storage connectors, so I added this as implementation to the base class instead.

python/hsfs/storage_connector.py
moritzmeister: This is a convenience method, not used anywhere. I know you are the opinion everything should go through On demand feature groups eventually.

python/hsfs/engine/spark.py
moritzmeister: true, refactored to same pattern

python/hsfs/storage_connector.py
moritzmeister: Okay, I know now what it's for, it prepares the options to connect with the `snowflake.connector` python library. the arguments are slightly different than the spark options. so it makes sense to have it. Added documentation with an exmple for it in the docstring

moritzmeister:
@davitbzh I think it's good if another pair of eyes has a look at this, so would be great if you can also give me a review.

I mostly tested JDBC, S3 and HOPSFS things, e.g. training datasets and on demand feature groups.

python/hsfs/storage_connector.py
```diff
@@ -336,302 +354,273 @@ def iam_role(self):
...
+        conn.prepare_spark()
+
+        spark.read.format("json").load("abfs://[bucket]/path")
+
+        # or
+        spark.read.format("json").load(conn.prepare_spark("abfs://[bucket]/path"))
+        ```
+
+        # Arguments
+            path: Path to prepare for reading from cloud storage. Defaults to `None`.
```
SirOibaf, line 510: The documentation here is wrong. See the `path` method above to for an example of path on abfss

python/hsfs/storage_connector.py
```diff
@@ -646,3 +635,105 @@ def snowflake_connector_options(self):
...
+        data_format: str = None,
+        options: dict = {},
+        path: str = None,
+    ):
+        """Reads a query into a dataframe using the storage connector."""
+        # options = (
+        #     {**self.spark_options(), **options}
+        #     if options is not None
+        #     else self.spark_options()
+        # )
```
SirOibaf, line 734: I guess this can be removed.

SirOibaf (changes requested):

python/hsfs/storage_connector.py
moritzmeister: I put it back and removed the single line, not sure why I commented it out

python/hsfs/storage_connector.py
moritzmeister: fixed

SirOibaf (approved):

PR #318, opened 2021-04-20
[HOPSWORKS-CLOUD-1822] Updated hopsfs permissions

smkniazi:


docs/hopsworksai/aws/restrictive_permissions.md
```diff
@@ -226,7 +226,6 @@ For this remove the following permissions from the instance profile:
       "S3:PutLifecycleConfiguration", 
       "S3:GetLifecycleConfiguration", 
       "S3:PutBucketVersioning", 
-      "S3:GetBucketVersioning"
```
berthoug, line 229: I'm surprised that it is the `GetBucketVersioning` and not the `PutBucketVersioning` that we remove. Why de we need to be able to put versioning?

smkniazi: the paragraph above talks about removing permissions that are not needed. now we need GetBucketVersioning for both enterprise and non-enterprise versions.

berthoug: Ok, I went too fast and thought it was the same as for azure.

berthoug (approved):

PR #323, opened 2021-04-22
[HOPSWORKS-2507] Offline JDBC fixes for Hudi external clients

SirOibaf:


moritzmeister (approved):

python/hsfs/client/external.py
```diff
@@ -109,10 +109,10 @@ def __init__(
                 self._cert_key = f.read()
 
             self._trust_store_path = _spark_session.conf.get(
-                "spark.hadoop.hops.ssl.keystore.name"
+                "spark.hadoop.hops.ssl.trustore.name"
```
jimdowling, line 112: Is this a typo?
spark.hadoop.hops.ssl.trustore.name
->
spark.hadoop.hops.ssl.truststore.name
See here
https://docs.hopsworks.ai/latest/integrations/spark/

moritzmeister: yes it is a typo but not here, here it's correct because it's already wrong in hopsworks

moritzmeister: but @SirOibaf knows better, can you double check fabio?

SirOibaf: Yeah it's a typo, but it originates from the hopsfs clients and it's spread a bit of everywhere now and spreads in the Hive Jars for Databricks, Hopsworks and here.

PR #334, opened 2021-04-28
[HOPSWORKS-CLOUD-1856] autoscale documentation

berthoug:
Fixes https://github.com/logicalclocks/hopsworks-cloud/issues/1856


docs/hopsworksai/autoscaling.md
```diff
@@ -0,0 +1,67 @@
...
+<p align="center">
+  <figure>
+    <a  href="../../assets/images/hopsworksai/configure_autoscale.png">
+      <img style="border: 1px solid #000" src="../../assets/images/hopsworksai/configure_autoscale.png" alt="Configure autoscale">
+    </a>
+    <figcaption>Configure autoscale</figcaption>
+  </figure>
+</p>
+
+Once you have clicked on *Configure autoscale* you will access a form allowing you to configure the autoscaling. This form is in two parts. In the first part, you configure the autoscaling for general-purpose compute nodes. In the second part, you configure the autoscaling for nodes equipped with GPUs. In both part you will have to set up the following:
```
kouzant, line 16:
```suggestion
Once you have clicked on *Configure autoscale* you will access a form allowing you to configure the autoscaling. This form is in two parts. In the first part, you configure the autoscaling for general-purpose compute nodes. In the second part, you configure the autoscaling for nodes equipped with GPUs. In both parts you will have to set up the following:
```

docs/hopsworksai/autoscaling.md
```diff
@@ -0,0 +1,67 @@
...
+
+1. The instance type you want to use. You can decide to not enable the autoscaling for GPU nodes by selecting *No GPU autoscale*.
+2. The size of the instances' disk.
+3. The minimum number of workers. 
+4. The maximum number of workers.
+5. The targeted number of overprovisioned nodes. Setting resource overprovisioning ensures that there are always some free resources in your cluster. This ensures that requests for new resources are fulfilled promptly. You configure the overprovisioning by setting the number of nodes you want to be overprovisioned. For example, if you set a value of *0.5* the system will start a new worker each time there are less than the resource equivalent of half a worker in the cluster. If you set this value to 0 new workers will only be started when a job or notebook request the resources.
```
kouzant, line 22:
```suggestion
5. The targeted number of overprovisioned nodes. Setting resource overprovisioning ensures that there are always some free resources in your cluster. This ensures that requests for new resources are fulfilled promptly. You configure the overprovisioning by setting the number of nodes you want to be overprovisioned. For example, if you set a value of *0.5* the system will start a new worker every time the aggregated cluster resources drop below 50% of a single worker's capabilities. If you set this value to 0 new workers will only be started when a job or notebook request the resources.
```

kouzant: I am not sure if my phrasing makes more sense, but this part needs some improvement

docs/hopsworksai/autoscaling.md
```diff
@@ -0,0 +1,67 @@
...
+6. The time to wait before removing unused resources. One often starts a new computation shortly after finishing the previous one. To avoid having to wait for workers to stop and start between each computation it is recommended to wait before shutting down workers. Here you set the amount of time in seconds resource need to be unused before they get removed from the system.
```
kouzant, line 23:
```suggestion
6. The time to wait before removing unused resources. One often starts a new computation shortly after finishing the previous one. To avoid having to wait for workers to stop and start between each computation it is recommended to wait before shutting down workers. Here you set the amount of time in seconds resources need to be unused before they get removed from the system.
```

docs/hopsworksai/autoscaling.md
```diff
@@ -0,0 +1,67 @@
...
+  <figure>
+    <a  href="../../assets/images/hopsworksai/configure_autoscale_review.png">
+      <img style="border: 1px solid #000" src="../../assets/images/hopsworksai/configure_autoscale_review.png" alt="Configure autoscale review">
+    </a>
+    <figcaption>Configure autoscale review</figcaption>
+  </figure>
+</p>
+
+## Modifying the autoscaling configuration
+You can update the autoscale configuration by going to the *Details* tab of the cluster and clicking on *Configure autoscale*. You will then go through the same steps as above. Note that if you change the *instance type*, nodes that are currently existing in the cluster with a different instance type will not be managed by the autoscale system anymore and you will have to remove them manually.
```
kouzant, line 55:
```suggestion
You can update the autoscale configuration by going to the *Details* tab of the cluster and clicking on *Configure autoscale*. You will then go through the same steps as above. Note that if you change the *instance type*, nodes that 
 currently exist in the cluster with a different instance type will not be managed by the autoscale system anymore and you will have to remove them manually. 
```

kouzant (changes requested):

kouzant (approved):

PR #361, opened 2021-05-19
[HOPSWORKS-2375] Prefix for joining to avoid clashing

davitbzh:


python/hsfs/constructor/query.py
```diff
@@ -148,12 +149,12 @@ def join(
                 the join. Defaults to `[]`.
             join_type: Type of join to perform, can be `"inner"`, `"outer"`, `"left"` or
                 `"right"`. Defaults to "inner".
-
+            prefix: User provided prefix to avoid feature name clash. Defaults to `None`.
```
moritzmeister, line 152: maybe explain that it will be applied to the "right" feature group that's being joined

moritzmeister (approved):

moritzmeister:
Just one thing, @davitbzh can you please test and create a training dataset from a query with a join from a pure Python Environment with the Hive engine?

PR #364, opened 2021-05-24
[Docs] Azure upgrade permissions required at resource group level not…

maismail:
… storage account

### DIFF
--- a/docs/hopsworksai/azure/getting_started.md
+++ b/docs/hopsworksai/azure/getting_started.md
@@ -247,10 +247,6 @@ Similarly to [Step 1.2](#step-12-creating-a-custom-role-for-hopsworksai) create 
             "Microsoft.Storage/storageAccounts/blobServices/containers/read",
             "Microsoft.Storage/storageAccounts/blobServices/write",
             "Microsoft.Storage/storageAccounts/blobServices/read",
-            "Microsoft.Compute/virtualMachines/read",
-            "Microsoft.Compute/virtualMachines/write",
-            "Microsoft.Compute/disks/read",
-            "Microsoft.Compute/disks/write"
         ],
         "notActions": [],
         "dataActions": [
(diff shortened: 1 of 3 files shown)


docs/hopsworksai/azure/upgrade.md
```diff
@@ -26,7 +26,7 @@ You need to **Stop** your cluster to start the upgrade process. Once your cluste
 ## Step 2: Add upgrade permissions to your user assigned managed identity
 
 !!! note
-    You can skip this step if you already have the following permissions in your user assigned managed identity:
+    You can skip this step if you already have the following permissions in your user assigned managed identity assigned to your resource group:
     ```json
```
berthoug, line 30: I don't understand what you are adding. The user `assigned managed identity` is not assigned to any resource group, it is assigned to the VMs, but I'm not sure it is helping.

docs/hopsworksai/azure/getting_started.md
```diff
@@ -352,6 +348,111 @@ Click on *Save* (7).
   </figure>
 </p>
 
+## Optional Step: Add upgrade permissions to your user assigned managed identity
```
berthoug, line 351: I'm not sure what to think about the changes made to this page. I think I get why you are trying to do. But, I would prefer the getting started to be as simple as possible and to have all the optional complexity described in more advanced part of the documentation.

berthoug (changes requested):

docs/hopsworksai/azure/getting_started.md
maismail: I agree in that case, i would just keep this change as part of the upgrade documentation instead

docs/hopsworksai/azure/upgrade.md
maismail: yeah, but we add role assignments with two different scopes one for storage account and for upgrade we need to add to resource group scope

berthoug: Then give names to the identity and put a link to where you create and give a name, I think it will be easier to read than what it is assigned to.

berthoug (approved):

PR #372, opened 2021-05-30
Fix validation for Python engine

SirOibaf:
When using the python engine, the validation should will be executed by the Hopsworks Job ingesting the data.


SirOibaf:
> In general, for the future, we should design these things with call to the engine, so that we don't need to have these `if`s about the engine type :)

Yes I had the same thought, however, I'd like to eventually implement data validation also for the python engine, in a similar way we do it for the feature group data ingestion.

PR #381, opened 2021-06-10
[HOPSWORKS-2606] compute training dataset statistics for train, test and validation splits separately

davitbzh:


SirOibaf:
Where is the python code?

java/src/main/java/com/logicalclocks/hsfs/TrainingDataset.java
```diff
@@ -319,7 +320,20 @@ public void show(int numRows) throws FeatureStoreException, IOException {
...
+        List<SplitStatistics> splitStatistics = new ArrayList<>();
+        for (Split split : this.splits) {
+          splitStatistics.add(new SplitStatistics(split.getName(),
+              statisticsEngine.computeStatistics(read(split.getName()),
+                  this.getStatisticsConfig().getColumns(),
+                  this.getStatisticsConfig().getHistograms(),
+                  this.getStatisticsConfig().getCorrelations(),
+                  null).getContent()));
+        }
+        return statisticsEngine.registerSplitStatistics(this, splitStatistics);
```
SirOibaf, line 333: This violates our convention of having all the computation and logic on the engines classes. I'd prefer if we don't break this convention as it worked quite well so far. I'd move the split iteration in the engine.

java/src/main/java/com/logicalclocks/hsfs/engine/TrainingDatasetEngine.java
```diff
@@ -108,12 +108,14 @@ public void save(TrainingDataset trainingDataset, Dataset<Row> dataset, Map<Stri
...
     trainingDataset.setVersion(apiTD.getVersion());
     trainingDataset.setId(apiTD.getId());
+    trainingDataset.setStorageConnector(apiTD.getStorageConnector());
 
     // Build write options map
     Map<String, String> writeOptions =
         SparkEngine.getInstance().getWriteOptions(userWriteOptions, trainingDataset.getDataFormat());
 
     SparkEngine.getInstance().write(trainingDataset, dataset, writeOptions, SaveMode.Overwrite);
+    return trainingDataset;
```
SirOibaf, line 118: The APIs should be consistent here between feature groups and training datasets. If you change the training dataset `save()` to return the object itself, then the feature group and on demand feature group classes should do the same.

python/hsfs/core/statistics_engine.py
```diff
@@ -75,6 +64,30 @@ def compute_statistics(
...
+            metadata_instance.statistics_config.columns,
+            metadata_instance.statistics_config.correlations,
+            metadata_instance.statistics_config.histograms,
+        )
+        return content_str
+
+    def register_split_statistics(self, td_metadata_instance, statistics_of_splits):
+        commit_time = int(float(datetime.datetime.now().timestamp()) * 1000)
+        stats = statistics.Statistics(
+            commit_time=commit_time, content=None, split_statistics=statistics_of_splits
```
SirOibaf, line 86: you don't need `content=None`, it's already the default value.

python/hsfs/training_dataset.py
```diff
@@ -248,7 +250,22 @@ def compute_statistics(self):
...
+                            self._statistics_engine.profile_statistics(
+                                self, self.read(split_name)
+                            ),
+                        )
+                    )
+                return self._statistics_engine.register_split_statistics(
+                    self, statistics_of_splits
+                )
+            else:
+                return self._statistics_engine.compute_statistics(self, self.read())
```
SirOibaf, line 268: Same comment as for the Java part. Let's be consistent with our policy of having all the business logic in the engines.

SirOibaf (changes requested):
What is this `python/hsfs/normalization.py` file - I guess it's a left over from another PR. Can you remove it?

java/src/main/java/com/logicalclocks/hsfs/TrainingDataset.java
tdoehmen: I just leave it here as a reminder -  when https://github.com/logicalclocks/feature-store-api/pull/407 gets merged, we have to pass the `exactUniqueness` parameter here as well.

python/hsfs/core/statistics_engine.py
tdoehmen: Same here -  when https://github.com/logicalclocks/feature-store-api/pull/407 gets merged, we have to also pass the `metadata_instance.statistics_config.exactUniqueness` to the `profile(...)` call.

SirOibaf (approved):

PR #382, opened 2021-06-14
[HOPSWORKS-2334] Add fallback to apikey file in Hopsworks client

javierdlrm:


python/hsfs/client/base.py
```diff
@@ -82,6 +83,11 @@ def _read_jwt(self):
         with open(os.path.join(self._secrets_dir, self.TOKEN_FILE), "r") as jwt:
             return jwt.read()
 
+    def _read_apikey(self):
```
moritzmeister, line 86: you could save the repetition of the code by making the `read_jwt` function general taking the filename as an argument. but it's not a deal breaker ;)

moritzmeister (approved):

PR #383, opened 2021-06-17
[HOPSWORKS-2617] add batch lookup API to retrieve batch of serving vectors from online feature store

davitbzh:


python/hsfs/core/training_dataset_engine.py
```diff
@@ -155,27 +155,71 @@ def deserialize_complex_features(self, feature_schemas, row_dict):
...
             self.init_prepared_statement(training_dataset, external)
 
+        prepared_statements = training_dataset.prepared_statements
+
+        _transaction = training_dataset.prepared_statement_connection.begin()
+        batch_executed_statements = [
+            self._execute_statements(training_dataset, entry, prepared_statements)
+            for entry in entries
+        ]
+        _transaction.commit()
```
moritzmeister, line 188: looks duplicated, that's handled better in the java part I believe

moritzmeister (changes requested):
I don't see where the auto commit is disabled in the JAVA api and where the transactions are started. Does it work differently with the Java JDBC api than with SQLalchemy or is it missing?

davitbzh:
> I don't see where the auto commit is disabled in the JAVA api and where the transactions are started. Does it work differently with the Java JDBC api than with SQLalchemy or is it missing?

it is disabled here: https://github.com/davitbzh/feature-store-api/blob/get_serving_vectors/java/src/main/java/com/logicalclocks/hsfs/engine/TrainingDatasetEngine.java#L207
and its started here https://github.com/davitbzh/feature-store-api/blob/get_serving_vectors/java/src/main/java/com/logicalclocks/hsfs/engine/TrainingDatasetEngine.java#L242
this is how I understood JAVA api is handling it

moritzmeister (approved):

PR #384, opened 2021-06-18
[DOCS] Hopsworks.ai terraform provider

maismail:


berthoug (commented):
Seems ok to me.
Maybe you can update the getting started step 2 to add a note saying "if you prefer terraform got there" and redirect to the appropriate example. This is if I get it well and the example does everything which is in step 2.

berthoug (approved):

PR #387, opened 2021-06-28
Better Job handling for Hive engine

SirOibaf:
* When calling save or insert on training datasets/feature groups
  return the object to the user so that they can monitor the execution

* Add an option (wait_for_job, default True) to wait for the execution
  to be finished before save or insert returns

* Update the documentation

### DIFF
--- a/python/hsfs/core/job_api.py
+++ b/python/hsfs/core/job_api.py
@@ -15,7 +15,7 @@
 #
 
 from hsfs import client
-from hsfs.core import job
+from hsfs.core import job, execution
 
 
 class JobApi:
@@ -35,3 +35,16 @@ class JobApi:
         path_params = ["project", _client._project_id, "jobs", name, "executions"]
 
         _client._send_request("POST", path_params)
+
+    def last_execution(self, job):
+        _client = client.get_instance()
+        path_params = ["project", _client._project_id, "jobs", job.name, "executions"]
+
+        query_params = {"limit": 1, "sort_by": "submissiontime:desc"}
+
+        headers = {"content-type": "application/json"}
+        return execution.Execution.from_response_json(
+            _client._send_request(
+                "GET", path_params, headers=headers, query_params=query_params
+            )
+        )
--- a/python/hsfs/engine/hive.py
+++ b/python/hsfs/engine/hive.py
@@ -253,19 +253,16 @@ class Engine:
 
         # Launch job
         print("Launching ingestion job...")
-        fg_job = self._job_api.launch(ingestion_job.job.name)
+        self._job_api.launch(ingestion_job.job.name)
         print(
             "Ingestion Job started successfully, you can follow the progress at {}".format(
                 self._get_job_url(ingestion_job.job.href)
             )
         )
 
-        # If the user passed the wait_for_job option consider it,
-        # otherwise use the default True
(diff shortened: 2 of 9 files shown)


python/hsfs/core/job_api.py
```diff
@@ -35,3 +35,16 @@ def launch(self, name):
         path_params = ["project", _client._project_id, "jobs", name, "executions"]
 
         _client._send_request("POST", path_params)
```
moritzmeister, line 37: shouldn't this return the job?

python/hsfs/engine/hive.py
```diff
@@ -249,45 +253,19 @@ def save_dataframe(
 
         # Launch job
         print("Launching ingestion job...")
-        self._job_api.launch(ingestion_job.job.name)
+        fg_job = self._job_api.launch(ingestion_job.job.name)
```
moritzmeister, line 256: this is where I think it breaks cause `job_api.launch()` doesn't return the job.

python/hsfs/engine/hive.py
```diff
@@ -249,45 +253,19 @@ def save_dataframe(
...
             "Ingestion Job started successfully, you can follow the progress at {}".format(
                 self._get_job_url(ingestion_job.job.href)
             )
         )
 
-    def _get_job_url(self, href: str):
-        """Use the endpoint returned by the API to construct the UI url for jobs
+        # If the user passed the wait_for_job option consider it,
+        # otherwise use the default True
+        if offline_write_options.pop("wait_for_job", True):
```
moritzmeister, line 265: you could move the `if` statement into the function and just use the boolean as condition in the while loop, so it's not repeated for training datasets and featuregroups

python/hsfs/training_dataset.py
```diff
@@ -194,20 +202,32 @@ def insert(
...
             `RestAPIError`: Unable to create training dataset metadata.
         """
-        self._training_dataset_engine.insert(self, features, write_options, overwrite)
+        # td_job is used only if the hive engine is used
+        td_job = self._training_dataset_engine.insert(
+            self, features, write_options, overwrite
+        )
+
+        if engine.get_type() == "spark":
+            self.compute_statistics()
```
moritzmeister, line 228: where does this come from?

moritzmeister (changes requested):

python/hsfs/core/job_api.py
SirOibaf: This actually returns an execution object not a job object.

python/hsfs/engine/hive.py
SirOibaf: True, I noticed that I already have the job object in `ingestion_job.job` so I'm using that one instead and reverting the changes on this line.

Also as mentioned in the other comment, `launch` returns an execution not a job

python/hsfs/engine/hive.py
SirOibaf: done.

python/hsfs/training_dataset.py
SirOibaf: I thought that it was a bug and the statistics computation was triggered with the `hive` engine. Actually the `compute_statistics()` contains already a check for the engine type.

moritzmeister (approved):

PR #394, opened 2021-07-06
Append features for Hudi feature groups

moritzmeister:


python/hsfs/engine/hive.py
```diff
@@ -340,6 +339,15 @@ def save_stream_dataframe(
             "Stream ingestion is not available on Python environments, because it requires Spark as engine."
         )
 
+    def get_empty_appended_dataframe(self, dataframe, new_features):
```
SirOibaf, line 342: Did you try to run it with the patch to enforce the feature group schema in the ingestion job? If it works, then we don't need this method.

SirOibaf (approved):

PR #395, opened 2021-07-06
[HOPSWORKS-2645] Support feature validation for on-demand FGs in hsfs

tkakantousis:


python/hsfs/core/feature_group_base_engine.py
```diff
@@ -52,3 +52,14 @@ def update_statistics_config(self, feature_group):
         self._feature_group_api.update_metadata(
             feature_group, feature_group, "updateStatsConfig"
         )
+
+    def update_config(self, feature_group, metadata):
+        """Update the metadata attribute specified of the feature group ."""
+        metadata_values = {
+            "updateStatsSettings": True,
+            "validationType": feature_group.validation_type,
+        }
```
moritzmeister, line 61: this looks like old code to me, I actually think we forgot to remove the function previously, cause look at line 50 above.

python/hsfs/core/feature_group_engine.py
```diff
@@ -23,7 +23,7 @@
 
 class FeatureGroupEngine(feature_group_base_engine.FeatureGroupBaseEngine):
     def save(self, feature_group, feature_dataframe, write_options):
-
+        print("FeatureGroupEngine.save()")
```
moritzmeister, line 26: do we really want these prints?

python/hsfs/core/feature_group_engine.py
```diff
@@ -154,17 +148,6 @@ def commit_delete(feature_group, delete_df, write_options):
         )
         return hudi_engine_instance.delete_record(delete_df, write_options)
 
-    def update_config(self, feature_group, metadata):
```
moritzmeister, line 157: I think this function was old and not actually used anymore, so you shouldn't need it in the base class. but please double check.

moritzmeister (changes requested):
You should be able to move the duplicated functions to the feature group base class.

python/hsfs/core/feature_group_base_engine.py
tkakantousis: It's still used here https://github.com/logicalclocks/feature-store-api/blob/ee70fecba586727d5ae9e5cc1d2e10463fad9819/python/hsfs/feature_group.py#L1224 

do you mean we should have an update_validation function?

java/src/main/java/com/logicalclocks/hsfs/OnDemandFeatureGroup.java
```diff
@@ -83,6 +89,12 @@ public OnDemandFeatureGroup(FeatureStore featureStore, @NonNull String name, Int
     this.storageConnector = storageConnector;
     this.features = features;
     this.statisticsConfig = statisticsConfig != null ? statisticsConfig : new StatisticsConfig();
+    this.validationType = validationType != null ? validationType : ValidationType.NONE;
+    if (expectations != null && !expectations.isEmpty()) {
```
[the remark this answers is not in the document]
tkakantousis: there is not Builder for the base class, which constructor would I add this attributes to?

java/src/main/java/com/logicalclocks/hsfs/engine/OnDemandFeatureGroupEngine.java
```diff
@@ -32,13 +33,18 @@
 
   public void saveFeatureGroup(OnDemandFeatureGroup onDemandFeatureGroup)
       throws FeatureStoreException, IOException {
+    Dataset<Row> onDemandDataset = null;
     if (onDemandFeatureGroup.getFeatures() == null) {
-      Dataset<Row> onDemandDataset = SparkEngine.getInstance()
+      onDemandDataset = SparkEngine.getInstance()
```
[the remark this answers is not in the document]
tkakantousis: By "this" do you mean registering the temp table? If so, I don't know, that's out of scope of this PR. That is currently on master https://github.com/logicalclocks/feature-store-api/blob/ce659adcfbc54a957da9f6ad041e950809f42285/java/src/main/java/com/logicalclocks/hsfs/engine/OnDemandFeatureGroupEngine.java#L35

this PR just adds the validation at the end of the saveFeatureGroup method

python/hsfs/storage_connector.py
```diff
@@ -705,7 +705,9 @@ def connection_string(self):
 
     @property
     def arguments(self):
-        """Additional JDBC, REDSHIFT, or Snowflake arguments."""
+        """Additional JDBC, REDSHIFT, or Snowflake arguments. When running hsfs with PySpark/Spark in Hopsworks,
```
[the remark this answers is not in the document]
tkakantousis: fixed

python/hsfs/core/feature_group_engine.py
tkakantousis: fixed

python/hsfs/core/feature_group_base_engine.py
tkakantousis: I will try to use the update_metadata function

java/src/main/java/com/logicalclocks/hsfs/metadata/FeatureGroupBase.java
```diff
@@ -80,8 +85,21 @@
   @Setter
   protected StatisticsConfig statisticsConfig = new StatisticsConfig();
 
+  @Getter
+  @Setter
+  @JsonProperty("validationType")
```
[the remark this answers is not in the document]
tkakantousis: fixed

python/hsfs/core/feature_group_base_engine.py
tkakantousis: fixed

python/hsfs/core/feature_group_engine.py
tkakantousis: fixed

java/src/main/java/com/logicalclocks/hsfs/OnDemandFeatureGroup.java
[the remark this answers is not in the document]
moritzmeister: you are right

moritzmeister (approved):

PR #399, opened 2021-07-16
[HOPSWORKS-2518] Point-In-Time Joins

moritzmeister:


moritzmeister:
@tkakantousis while merging the upstream branch I had some conflicts regarding validation, would be great if you could also have a look

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
```diff
@@ -165,8 +166,11 @@ private FeatureGroupCommit getLastCommitMetadata(SparkSession sparkSession, Stri
     hudiArgs.put(HUDI_WRITE_INSERT_DROP_DUPLICATES, "true");
 
     // primary keys
-    Seq<String> primaryColumns = utils.getPrimaryColumns(featureGroup);
-    hudiArgs.put(HUDI_RECORD_KEY, primaryColumns.mkString(","));
+    String primaryColumns = utils.getPrimaryColumns(featureGroup).mkString(",");
+    if (!Strings.isNullOrEmpty(featureGroup.getEventTime())) {
+      primaryColumns = primaryColumns + "," + featureGroup.getEventTime();
```
SirOibaf, line 171: This looks a bit weird, why not put it in the collection just have a single string manipulation?

python/hsfs/feature_store.py
```diff
@@ -354,6 +355,9 @@ def create_feature_group(
             expectations: Optionally, a list of expectations to be attached to the feature group.
                 The expectations list contains Expectation metadata objects which can be retrieved with
                 the `get_expectation()` and `get_expectations()` functions.
+            event_time: Optionally, provide the name of the feature containing the event
```
SirOibaf, line 358: I think we should have a separate Guide page on the documentation explaining PIT joins.

SirOibaf (approved):

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
moritzmeister: I agree, and I am sure I tried, but the collection is a Scala Seq, and I could not manage adding an element using scala syntax from java.

tkakantousis (approved):

jimdowling:
Can we get this merged?
It comes up a lot in demos and i have nothing to show prospects.

PR #400, opened 2021-07-19
[HOPSWORKS-2680] Save code (jupyter/job) that it's used to create a feature group

bubriks:


python/hsfs/core/code_engine.py
```diff
@@ -0,0 +1,39 @@
...
+import os
+
+
+class CodeEngine:
+    def __init__(self, feature_store_id, entity_type):
+        self._code_api = code_api.CodeApi(
+            feature_store_id, entity_type
+        )
+
+    def compute_code(
```
SirOibaf, line 30: Same comment about the naming of the method

python/hsfs/core/code_engine.py
```diff
@@ -0,0 +1,39 @@
...
+        self, metadata_instance, feature_dataframe=None, feature_group_commit_id=None
+    ):
+        """Compute code for a dataframe and send the result json to Hopsworks."""
+        code_entity = code.Code(
+            commit_time=int(float(datetime.datetime.now().timestamp()) * 1000),
+            feature_group_commit_id=feature_group_commit_id,
```
SirOibaf, line 36: `feature_group_commit_id` is always null. What happens in the backend?

python/hsfs/core/code_engine.py
```diff
@@ -0,0 +1,39 @@
...
+            application_id=os.environ["APPLICATION_WEB_PROXY_BASE"][7:],
+        )
+        self._code_api.post(metadata_instance, code_entity, os.environ["HOPSWORKS_KERNEL_ID"], "JUPYTER")
```
SirOibaf, line 39: In Python, if an environment variable doesn't exist, an exception is thrown. Here you need to be careful. As I mentioned above, you can run HSFS as part of a Job for instance, which doesn't have the `HOPSWORKS_KERNEL_ID` env variable.

java/src/main/java/com/logicalclocks/hsfs/FeatureGroup.java
```diff
@@ -222,6 +224,7 @@ public void save(Dataset<Row> featureData, Map<String, String> writeOptions)
       throws FeatureStoreException, IOException, ParseException {
     featureGroupEngine.save(this, featureData, primaryKeys, partitionKeys, hudiPrecombineKey,
         writeOptions);
+    codeEngine.computeCode(this, null);
```
SirOibaf, line 227: You need to add these methods also to the `OnDemandFeatureGroup` class.

python/hsfs/code.py
```diff
@@ -0,0 +1,75 @@
...
+        self._content = "" if content is None else json.loads(content)
+
+    @classmethod
+    def from_response_json(cls, json_dict):
+        json_decamelized = humps.decamelize(json_dict)
+        # Currently getting multiple commits at the same time is not allowed
+        if json_decamelized["count"] == 0:
+            return None
+        elif len(json_decamelized["items"]) == 1:
+            return cls(**json_decamelized["items"][0])
```
SirOibaf, line 48: This method is not used as we don't support getting the code from the Hopsworks API, you can remove it.

python/hsfs/code.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def to_dict(self):
+        return {
+            "commitTime": self._commit_time,
+            "applicationId": self._application_id,
+            "featureGroupCommitId": self._feature_group_commit_id,
+            "content": json.dumps(self._content),
```
SirOibaf, line 55: `content` is always going to be null so we can remove it from here.

python/hsfs/code.py
```diff
@@ -0,0 +1,75 @@
...
+        href=None,
+        expand=None,
+        items=None,
+        count=None,
+        type=None,
+    ):
+        self._commit_time = commit_time
+        self._application_id = application_id
+        self._feature_group_commit_id = feature_group_commit_id
+        self._content = "" if content is None else json.loads(content)
```
SirOibaf, line 39: `content` is always going to be null so we can remove it from here.

python/hsfs/feature_group.py
```diff
@@ -705,6 +709,7 @@ def save(
 
         # fg_job is used only if the hive engine is used
         fg_job = self._feature_group_engine.save(self, feature_dataframe, write_options)
+        self._code_engine.compute_code(self)
```
SirOibaf, line 712: You need to add these calls also on the `OnDemandFeatureGroup` class - for the save and insert method.

SirOibaf (changes requested):

python/hsfs/feature_group.py
bubriks: OnDemandFeatureGroup doesn't have insert, only save.

java/src/main/java/com/logicalclocks/hsfs/engine/CodeEngine.java
```diff
@@ -0,0 +1,79 @@
...
+
+import java.io.IOException;
+import java.sql.Timestamp;
+import java.time.LocalDateTime;
+
+public class CodeEngine {
+
+  private CodeApi codeApi;
+  private static String kernelEnv = "HOPSWORKS_KERNEL_ID";
+  private static String webProxyEnv = "APPLICATION_WEB_PROXY_BASE";
```
SirOibaf, line 36: Java convention dictates that static variables should be named with all capital letters. Also add final (`private static final`)

java/src/main/java/com/logicalclocks/hsfs/engine/CodeEngine.java
```diff
@@ -0,0 +1,79 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(CodeEngine.class);
+
+  public CodeEngine(EntityEndpointType entityType) {
+    this.codeApi = new CodeApi(entityType);
+  }
+
+  public Code saveCode(TrainingDataset trainingDataset)
+          throws FeatureStoreException, IOException {
+    String kernelId = System.getenv(kernelEnv);
+    if(kernelId == null || kernelId.isEmpty()){
```
SirOibaf, line 47: this code doesn't pass the checkstyle.

java/src/main/java/com/logicalclocks/hsfs/engine/CodeEngine.java
```diff
@@ -0,0 +1,79 @@
...
+      return null;
+    }
+    return codeApi.post(trainingDataset, saveCode(),
+            kernelId, RunType.JUPYTER.toString());
```
SirOibaf, line 51: Is better if you pass around the enum and let the Jackson library convert it to string. It makes the code safer and more readable.

java/src/main/java/com/logicalclocks/hsfs/engine/CodeEngine.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  public Code saveCode(FeatureGroupBase featureGroup)
+          throws FeatureStoreException, IOException {
+    String kernelId = System.getenv(kernelEnv);
+    if(kernelId == null || kernelId.isEmpty()){
```
SirOibaf, line 57: Use `Strings.isNullOrEmpty()` to check - it's more consistent with the rest of the codebase. (use the one from the google package)

java/src/main/java/com/logicalclocks/hsfs/engine/CodeEngine.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  private Code saveCode() {
+    Long commitTime = Timestamp.valueOf(LocalDateTime.now()).getTime();
+    String applicationId = null;
+    String webProxy = System.getenv(webProxyEnv);
+    if(webProxy != null && !webProxy.isEmpty()){
+      applicationId = webProxy.substring(7);
+    }
+    return new Code(commitTime, null, applicationId, "");
```
SirOibaf, line 71: Since you are not providing the `featureGroupCommitId` and `code` fields, please create a constructor without them and use the new constructor instead.

python/hsfs/core/code_engine.py
```diff
@@ -0,0 +1,50 @@
...
+            feature_store_id, entity_type
+        )
+
+    def save_code(self, metadata_instance):
+        """Compute code for a dataframe and send the result json to Hopsworks."""
+        kernel_id = os.environ.get(CodeEngine.KERNEL_ENV)
+        if not kernel_id:
+            return
+        web_proxy = os.environ.get(CodeEngine.WEB_PROXY_ENV)
+        application_id = None if web_proxy is None else web_proxy[7:]
```
SirOibaf, line 40: This statement can be simplified:

application_id =  web_proxy[7:] if web_proxy else None

You can also do the operation directly in the `code.Code` constructor instead of defining a temporary variable.

SirOibaf (changes requested):
Looks good, just some minor style comment. Also check the Java compilation, it fails the checkstyle.

SirOibaf (approved):

PR #401, opened 2021-07-19
[HOPSWORKS-2657] Use correct Transformation Function DTO in JAVA API

davitbzh:


java/src/main/java/com/logicalclocks/hsfs/metadata/TrainingDatasetApi.java
```diff
@@ -197,6 +197,8 @@ public TrainingDataset getTransformationFunctions(TrainingDataset trainingDatase
         .expand();
 
     LOGGER.info("Sending metadata request: " + uri);
-    return hopsworksClient.handleRequest(new HttpGet(uri), TrainingDataset.class);
+    TransformationFunctionAttached transformationFunctionAttached =
+        hopsworksClient.handleRequest(new HttpGet(uri), TransformationFunctionAttached.class);
+    return transformationFunctionAttached.getItems();
```
SirOibaf, line 202: I believe that if the training dataset doesn't have any transformation functions attached, then `items()` is null. This will return a null to the caller `TrainingDatasetEngine` which then calls `.size()` throwing a NullPointerException.

Can you verify that please?

SirOibaf (approved):

PR #418, opened 2021-08-05
[HOPSWORKS-2690] Allow transformation functions to produce new, transformed column when creating training dataset

davitbzh:


python/hsfs/transformation_function.py
```diff
@@ -17,17 +17,11 @@
 import inspect
 import ast
 
-from hsfs import util, engine
+from hsfs import util
 from hsfs.core import transformation_function_engine
 
 
 class TransformationFunction:
-    UDF = "@udf"
```
moritzmeister, line 25: why are these not needed anymore?

python/hsfs/engine/spark.py
```diff
@@ -315,14 +315,41 @@ def write_training_dataset(
...
-            dataset = dataset.withColumn(
-                feature_name, transformation_fn.transformation_fn(col(feature_name))
+            fn_registration_name = (
+                transformation_fn.name + "_" + str(transformation_fn.version)
             )
+            self._spark_session.udf.register(
+                fn_registration_name, transformation_fn.transformation_fn
+            )
+            transformation_fn_expressions.append(
+                "{fn_name:}({name:}) AS {name:}".format(
```
moritzmeister, line 332: why do you need the `:` I think they add some formatting spaces in some cases, is that wanted?

moritzmeister (commented):
general questions

python/hsfs/transformation_function.py
davitbzh: No its not necessary any more. We use `self._spark_session.udf.register` for `selectExpr`

python/hsfs/engine/spark.py
davitbzh: yeah, it doesn't make any difference. It seemed better readable for my eyes :)

moritzmeister (approved):

PR #423, opened 2021-08-09
[HOPSWORKS-2598] Add application field in snowflake storage connector

bubriks:


java/src/main/java/com/logicalclocks/hsfs/StorageConnector.java
```diff
@@ -294,6 +294,9 @@ public String getPath(String subPath) {
     @Getter @Setter
     private String table;
 
+    @Getter @Setter
+    private String application;
```
SirOibaf, line 298: The application field should also be added to the map returned by the `sparkOptions()` method, similar to what you did for the Python version.

SirOibaf (changes requested):

python/hsfs/storage_connector.py
```diff
@@ -652,6 +661,8 @@ def spark_options(self):
             props["sfToken"] = self._token
         if self._warehouse is not None:
             props["sfWarehouse"] = self._warehouse
+        if self._application is not None:
```
SirOibaf, line 664: keep just `if self._application:` - in python empty string evaluates as false, and we don't want to set the field if the string is None or empty

SirOibaf (approved):

PR #425, opened 2021-08-10
[HOPSWORKS-2695] Replace creator with UserDTO in FeaturestoreEntityDTO

bubriks:


java/src/main/java/com/logicalclocks/hsfs/metadata/User.java
```diff
@@ -0,0 +1,93 @@
...
+ */
+
+package com.logicalclocks.hsfs.metadata;
+
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@NoArgsConstructor
+public class User {
```
SirOibaf, line 24: You don't need to add all the fields here. First the backend doesn't populate most of them. Second you can add `@JsonIgnoreProperties(ignoreUnknown = true)` as class annotation to not make it fail in the case we ever change the backend.

python/hsfs/feature_group.py
```diff
@@ -468,7 +468,7 @@ def __init__(
         self._feature_store_name = featurestore_name
         self._description = description
         self._created = created
-        self._creator = creator
+        self._creator = user.User.from_response_json(creator)
```
SirOibaf, line 471: Did you test this? the `creator` field is available only if the data comes from the backend. If the class is initialized by the user through the API `creator` is going to be null.

If you see the code below the one you changed, you see that we differentiate between the case in which the class is instantiated with data coming from the backend or data added by the user.

SirOibaf (changes requested):

python/hsfs/feature_group.py
bubriks: if creator is null, then result of user.User.from_response_json(creator) is also None

bubriks: here is an example:

bubriks: image

bubriks: i could change it as follows, if it is still needed:
if creator is not None:
    self._creator = user.User.from_response_json(creator)
else:
    self._creator = None

SirOibaf (approved):

PR #430, opened 2021-08-16
Validate Spark configuration when calling connect()

bubriks:


java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java
```diff
@@ -93,6 +95,53 @@ private SparkEngine() {
     sparkSession.conf().set("hive.exec.dynamic.partition.mode", "nonstrict");
     // force Spark to fallback to using the Hive Serde to read Hudi COPY_ON_WRITE tables
     sparkSession.conf().set("spark.sql.hive.convertMetastoreParquet", "false");
+
+    validateSparkConfiguration();
+  }
+
+  @SneakyThrows
```
moritzmeister, line 102: why did you choose not to declare the `throws`?

java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java
```diff
@@ -93,6 +95,53 @@ private SparkEngine() {
...
+  private void validateSparkConfiguration() {
+    String key = "spark.hadoop.hops.ssl.trustore.name";
+    if (sparkSession.conf().get(key).isEmpty())
+      throw new FeatureStoreException("Invalid: " + key);
```
moritzmeister, line 106: Maybe make the message a bit more informative and define the repeated part only once, e.g.
`"Spark is misconfigured for communication with Hopsworks, missing or invalid property: xyz"`

java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java
```diff
@@ -93,6 +95,53 @@ private SparkEngine() {
...
+    key = "spark.hadoop.fs.hopsfs.impl";
+    if (!sparkSession.conf().get(key)
+            .equals("io.hops.hopsfs.client.HopsFileSystem"))
+      throw new FeatureStoreException("Invalid: " + key);
+    key = "spark.hadoop.hops.ssl.keystores.passwd.name";
+    if (sparkSession.conf().get(key).isEmpty())
+      throw new FeatureStoreException("Invalid: " + key);
+    key = "spark.hadoop.hops.ipc.server.ssl.enabled";
+    if (!sparkSession.conf().get(key)
+            .equals("true"))
```
moritzmeister, line 131: nitpick, should fit on one line

java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java
```diff
@@ -93,6 +95,53 @@ private SparkEngine() {
...
+      throw new FeatureStoreException("Invalid: " + key);
+    //todo should jars path be compared?
```
moritzmeister, line 133: No I think it shouldn't, on databricks we set it, but users might use their own spark setup and have it in some other location.

java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java
```diff
@@ -93,6 +95,53 @@ private SparkEngine() {
...
+    key = "spark.sql.hive.metastore.jars";
+    if (!sparkSession.conf().get(key)
+            .equals("/hopsworks_metastore_jar/lib/*"))
+      throw new FeatureStoreException("Invalid: " + key);
+    key = "spark.hadoop.client.rpc.ssl.enabled.protocol";
+    if (!sparkSession.conf().get(key)
+            .equals("TLSv1.2"))
```
moritzmeister, line 140: possibly one line?

python/hsfs/engine/spark.py
```diff
@@ -53,10 +53,48 @@ def __init__(self):
...
 
+        self.validate_spark_configuration()
+
         if importlib.util.find_spec("pydoop"):
             # If we are on Databricks don't setup Pydoop as it's not available and cannot be easily installed.
             util.setup_pydoop()
 
+    def validate_spark_configuration(self):
+        key = "spark.hadoop.hops.ssl.trustore.name"
+        if(not self._spark_session.conf.get(key)):
```
moritzmeister, line 64: no need for parenthesis in python:
```suggestion
        if not self._spark_session.conf.get(key):
```

python/hsfs/engine/spark.py
```diff
@@ -53,10 +53,48 @@ def __init__(self):
...
+            raise FeatureStoreException("Invalid: " + key)
+        key = "spark.hadoop.hops.rpc.socket.factory.class.default"
+        if(not self._spark_session.conf.get(key) == "io.hops.hadoop.shaded.org.apache.hadoop.net.HopsSSLSocketFactory"):
```
moritzmeister, line 67: I am not sure what the data type of `conf` actually is, but I would think it is a `dict` so are you sure it doesn't throw a `KeyError` when the setting doesn't exist at all? This might need to be handled for properties which are our own, and are not available at all on other environments.

moritzmeister (changes requested):

java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java
bubriks: because it would require added "throws FeatureStoreException" to quite a few methods and thought that it would be cleaner like this.

java/src/main/java/com/logicalclocks/hsfs/HopsworksConnection.java
```diff
@@ -89,6 +90,8 @@ public HopsworksConnection(String host, int port, String project, Region region,
     this.apiKeyFilePath = apiKeyFilePath;
     this.apiKeyValue = apiKeyValue;
 
+    SparkEngine.getInstance();
```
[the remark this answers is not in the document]
bubriks: otherwise configuration was not checked on connection

moritzmeister (approved):

moritzmeister (approved):

PR #460, opened 2021-09-07
Validate spark configuration

bubriks:


python/hsfs/client/external.py
```diff
@@ -114,6 +115,44 @@ def __init__(
...
+            raise FeatureStoreException(exception_text + key)
+        key = "spark.sql.hive.metastore.jars"
+        if not _spark_session.conf.get(key, None):
+            raise FeatureStoreException(exception_text + key)
+        key = "spark.hadoop.client.rpc.ssl.enabled.protocol"
+        if not _spark_session.conf.get(key, None) == "TLSv1.2":
+            raise FeatureStoreException(exception_text + key)
+        key = "spark.hadoop.hive.metastore.uris"
+        if not _spark_session.conf.get(key, None):
+            raise FeatureStoreException(exception_text + key)
```
SirOibaf, line 155: We should by DRY there. Define a list of keys you need to check and then iterate over the list instead of repeating the code.

python/hsfs/client/external.py
```diff
@@ -114,6 +115,44 @@ def __init__(
             self._key_store_path = _spark_session.conf.get(
                 "spark.hadoop.hops.ssl.keystore.name"
             )
+            self.validate_spark_configuration(_spark_session)
```
SirOibaf, line 118: The validation should happen before we try to access the properties in the line above.

java/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksExternalClient.java
```diff
@@ -141,6 +141,7 @@ public HopsworksExternalClient(CloseableHttpClient httpClient, HttpHost httpHost
     this.trustStorePath = SparkEngine.getInstance().getTrustStorePath();
     this.keyStorePath = SparkEngine.getInstance().getKeyStorePath();
     this.certKey = HopsworksHttpClient.readCertKey(SparkEngine.getInstance().getCertKey());
+    SparkEngine.getInstance().validateSparkConfiguration();
```
SirOibaf, line 144: Validation should happen before we try to access the properties.

java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java
```diff
@@ -95,6 +95,56 @@ private SparkEngine() {
...
+      throw new FeatureStoreException(exceptionText + key);
+    }
+    key = "spark.hadoop.client.rpc.ssl.enabled.protocol";
+    if (!sparkSession.conf().get(key, "").equals("TLSv1.2")) {
+      throw new FeatureStoreException(exceptionText + key);
+    }
+    key = "spark.hadoop.hive.metastore.uris";
+    if (sparkSession.conf().get(key, "").isEmpty()) {
+      throw new FeatureStoreException(exceptionText + key);
+    }
```
SirOibaf, line 145: Same comment on being DRY - Define a list and iterate over it.

SirOibaf (changes requested):

java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java
```diff
@@ -95,6 +95,36 @@ private SparkEngine() {
...
+        if (sparkSession.conf().get(entry.getKey(), "").isEmpty()) {
+          throw new FeatureStoreException(exceptionText + entry.getKey());
+        }
+      } else {
+        if (!sparkSession.conf().get(entry.getKey(), "").equals(entry.getValue())) {
+          throw new FeatureStoreException(exceptionText + entry.getKey());
+        }
+      }
+    }
+  }
```
SirOibaf, line 126: Here the code can be simplified, also if the entry.value is null, you can use `contains` on the map, instead of doing the string comparison.

```suggestion
    for (Map.Entry<String, String> entry : configurationMap.entrySet()) {
      if (entry.getValue() == null && !sparkSession.conf().contains(entry.getKey())) {
        throw new FeatureStoreException(exceptionText + entry.getKey());
      } else if (!sparkSession.conf().get(entry.getKey(), "").equalsIgnoreCase(entry.getValue())) {
        throw new FeatureStoreException(exceptionText + entry.getKey());
      }
    }
```

python/hsfs/client/external.py
```diff
@@ -115,6 +117,29 @@ def __init__(
...
+                "spark.hadoop.hops.rpc.socket.factory.class.default": "io.hops.hadoop.shaded.org.apache.hadoop.net.HopsSSLSocketFactory",
+                "spark.serializer": "org.apache.spark.serializer.KryoSerializer",
+                "spark.hadoop.hops.ssl.hostname.verifier": "ALLOW_ALL",
+                "spark.hadoop.hops.ssl.keystore.name": None,
+                "spark.hadoop.fs.hopsfs.impl": "io.hops.hopsfs.client.HopsFileSystem",
+                "spark.hadoop.hops.ssl.keystores.passwd.name": None,
+                "spark.hadoop.hops.ipc.server.ssl.enabled": "true",
+                "spark.sql.hive.metastore.jars": None,
+                "spark.hadoop.client.rpc.ssl.enabled.protocol": "TLSv1.2",
+                "spark.hadoop.hive.metastore.uris": None}
```
SirOibaf, line 133: This looks weirdly formatted - did you use `black`?

python/hsfs/client/external.py
```diff
@@ -115,6 +117,29 @@ def __init__(
...
+        
+        for key, value in configuration_dict.items():
+            if value == None:
+                if not _spark_session.conf.get(key, None):
```
SirOibaf, line 137: You can avoid the nested ifs here, just use `and`

SirOibaf (changes requested):

python/hsfs/client/external.py
bubriks: Sorry i don't understand what you mean here.

SirOibaf (approved):

PR #466, opened 2021-09-10
Add primary key info to on demand feature groups

bubriks:


python/hsfs/feature_store.py
```diff
@@ -415,6 +416,11 @@ def create_on_demand_feature_group(
             description: A string describing the contents of the on-demand feature group to
                 improve discoverability for Data Scientists, defaults to empty string
                 `""`.
+            primary_key: A list of feature names to be used as primary key for the
+                feature group. This primary key can be a composite key of multiple
+                features and will be used as joining key, if not specified otherwise.
+                Defaults to empty list `[]`, and the first column of the DataFrame will
+                be used as primary key.
```
SirOibaf, line 423: The documentation here is wrong. If you don't provide any primary key, then the feature group won't have any primary key at all. 

I believe you have taken the documentation from the `create_feature_group` method. The documentation is wrong there as well. Can you please fix it in both places?

python/hsfs/feature_group.py
```diff
@@ -1289,6 +1290,9 @@ def __init__(
 
         if self._id:
             # Got from Hopsworks, deserialize features and storage connector
+            self.primary_key = [
+                feat.name for feat in self._features if feat.primary is True
+            ]
```
SirOibaf, line 1295: This doesn't work. `self._feature` is empty until the next line.

SirOibaf: Indeed, if you try to do `fs.get_on_demand_feature_group("name_of_fg")` it fails.

java/src/main/java/com/logicalclocks/hsfs/OnDemandFeatureGroup.java
```diff
@@ -108,7 +110,7 @@ public OnDemandFeatureGroup(FeatureStore featureStore, int id) {
   }
 
   public void save() throws FeatureStoreException, IOException {
-    onDemandFeatureGroupEngine.saveFeatureGroup(this);
+    onDemandFeatureGroupEngine.saveFeatureGroup(this, primaryKeys);
```
SirOibaf, line 113: You don't need to provide the `primaryKeys` here, they are already set in the object.

SirOibaf (changes requested):

SirOibaf:
Formatting is still not good, I'm not sure what you are doing, but `black8 --check` still fails on this PR. I'll merge it and fix it in another one.

SirOibaf (approved):

PR #469, opened 2021-09-15
[HOPSWORKS-2732] Add github actions for building documentation

robzor92:


.github/workflows/mkdocs-master.yml
```diff
@@ -0,0 +1,38 @@
+name: mkdocs
```
moritzmeister, line 1: You might need to give them unique names here, so we can distinguish them in the Github UI where the runs are shown.
```suggestion
name: mkdocs-master
```

.github/workflows/mkdocs-release.yml
```diff
@@ -0,0 +1,38 @@
+name: mkdocs
```
moritzmeister, line 1: Same as above.
```suggestion
name: mkdocs-release
```

moritzmeister (changes requested):

.github/workflows/mkdocs-master.yml
robzor92: I set unique names on all PRs

.github/workflows/mkdocs-release.yml
robzor92: Done

moritzmeister (approved):

PR #472, opened 2021-09-15
API calls for getting all versions of entities

moritzmeister:


java/src/main/java/com/logicalclocks/hsfs/metadata/TrainingDatasetApi.java
```diff
@@ -49,6 +50,31 @@
...
+        .expand();
+
+    LOGGER.info("Sending metadata request: " + uri);
+    TrainingDataset[] trainingDatasets = hopsworksClient.handleRequest(new HttpGet(uri), TrainingDataset[].class);
+
+    for (TrainingDataset td : trainingDatasets) {
+      td.setFeatureStore(featureStore);
+      td.getFeatures().stream()
+          .filter(f -> f.getFeaturegroup() != null)
+          .forEach(f -> f.getFeaturegroup().setFeatureStore(featureStore));
```
SirOibaf, line 73: I'm confused on why we need this - I see that the other get has it as well. It doesn't look correct though - what happens if you are joining features coming from different feature groups?

SirOibaf: Also I think we should be DRY here and apply the same logic as we do for the feature groups, so we don't have twice the same get method.

SirOibaf (changes requested):

java/src/main/java/com/logicalclocks/hsfs/metadata/TrainingDatasetApi.java
moritzmeister: The setting of the feature store object on the feature group was introduced when we added the OnlineFS ingestion, I think it is needed to get the schema for complex features when decoding them in the `get_serving_vector()`. I am not sure why we don't have that in the Python API though.
We are only setting the feature store, not the feature group itself, all features in a training dataset should be coming from the same feature store right?

moritzmeister: I agree that I can make it dry

moritzmeister: okay in python we have set up the API classes a bit differently and we keep a reference to the feature store there, that's why we don't do it there.

moritzmeister:
#473 remember this for future. We set the same feature store for all features in a training dataset. In python we cache it in the API classes which is also incorrect.

java/src/main/java/com/logicalclocks/hsfs/metadata/TrainingDatasetApi.java
moritzmeister: see #473

SirOibaf (approved):

PR #480, opened 2021-09-27
Fix validate spark configuration

bubriks:


python/hsfs/client/external.py
```diff
@@ -133,11 +133,11 @@ def validate_spark_configuration(self, _spark_session):
         }
 
         for key, value in configuration_dict.items():
-            if value is None and not _spark_session.conf.get(key, None):
+            if not (
+                key in _spark_session.conf.keys()
```
moritzmeister, line 137: see my messages on slack.

moritzmeister (changes requested):

moritzmeister (approved):

PR #518, opened 2021-11-02
Add method to rebuild query logic/filter options from a json object

SirOibaf:
This is used when rebuilding the query with filters from a JSON object
when using the UI or the Python API to generate a training dataset

### DIFF
--- a/python/hsfs/constructor/filter.py
+++ b/python/hsfs/constructor/filter.py
@@ -107,8 +107,8 @@ class Logic:
             type=json_dict["type"] if "type" in json_dict else None,
             left_f=json_dict["left_filter"] if "left_filter" in json_dict else None,
             right_f=json_dict["right_filter"] if "right_filter" in json_dict else None,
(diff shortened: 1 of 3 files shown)


moritzmeister (approved):

python/hsfs/constructor/filter.py
```diff
@@ -98,6 +98,19 @@ def to_dict(self):
...
 
+    @classmethod
+    def from_response_json(cls, json_dict):
+        if json_dict is None:
+            return None
+
+        return cls(
+            type=json_dict["type"] if "type" in json_dict else None,
+            left_f=json_dict["left_filter"] if "left_filter" in json_dict else None,
+            right_f=json_dict["right_filter"] if "right_filter" in json_dict else None,
```
moritzmeister, line 109: actually shouldn't this be also from_response_json?

moritzmeister (approved):

PR #530, opened 2021-11-19
[HOPSWORKS-2838] Training dataset get_query for offline storage not working on hive engine

### LINKED ISSUE #523: Training dataset get_query for offline storage not working on hive engine
When we send a QueryDTO during training dataset creation we specifically set a flag in the DTO for hive engine.

However, the QueryDTO is not used when calling `td.get_query(online=False)` because the query is just retrieved from the dtabase. in this case from the hive engine a query string for Spark is returned but it should be for hive engine.

To circumvent this we need to add a query param to the query endpoint for training datasets.

kennethmhc:
resolve #523 

## Changes
1. Add "hive query" parameter to java and python client.

### DIFF
--- a/java/src/main/java/com/logicalclocks/hsfs/TrainingDataset.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/TrainingDataset.java
@@ -423,28 +423,22 @@ public class TrainingDataset {
 
   @JsonIgnore
   public String getQuery() throws FeatureStoreException, IOException {
-    return getQuery(Storage.ONLINE, false, false);
+    return getQuery(Storage.ONLINE, false);
   }
 
   @JsonIgnore
   public String getQuery(boolean withLabel) throws FeatureStoreException, IOException {
-    return getQuery(Storage.ONLINE, withLabel, false);
+    return getQuery(Storage.ONLINE, withLabel);
   }
 
   @JsonIgnore
   public String getQuery(Storage storage) throws FeatureStoreException, IOException {
-    return getQuery(storage, false, false);
+    return getQuery(storage, false);
   }
   
   @JsonIgnore
   public String getQuery(Storage storage, boolean withLabel) throws FeatureStoreException, IOException {
     return trainingDatasetEngine.getQuery(this, storage, withLabel, false);
-  }
-
-  @JsonIgnore
-  public String getQuery(Storage storage, boolean withLabel, boolean isHiveQuery)
-      throws FeatureStoreException, IOException {
-    return trainingDatasetEngine.getQuery(this, storage, withLabel, isHiveQuery);
   }
 
   @JsonIgnore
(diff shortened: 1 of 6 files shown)


python/hsfs/training_dataset.py
```diff
@@ -619,7 +619,9 @@ def query(self):
         """Query to generate this training dataset from online feature store."""
         return self._training_dataset_engine.query(self, True, True, False)
 
-    def get_query(self, online: bool = True, with_label: bool = False, is_hive_query: bool = False):
+    def get_query(
+        self, online: bool = True, with_label: bool = False, is_hive_query: bool = False
```
moritzmeister, line 623: We don't actually want to surface this to the user, the user is not necessarily aware of what engine he is using since it's all transparent. E.g. it will use Spark if available otherwise not.

So you can get the engine type from the engine singleton like this:
```
from hsfs import engine
engine.get_type()
```

java/src/main/java/com/logicalclocks/hsfs/TrainingDataset.java
```diff
@@ -423,22 +423,28 @@ public void deleteTag(String name) throws FeatureStoreException, IOException {
...
-
+  
   @JsonIgnore
   public String getQuery(Storage storage, boolean withLabel) throws FeatureStoreException, IOException {
-    return trainingDatasetEngine.getQuery(this, storage, withLabel);
+    return trainingDatasetEngine.getQuery(this, storage, withLabel, false);
+  }
+
+  @JsonIgnore
+  public String getQuery(Storage storage, boolean withLabel, boolean isHiveQuery)
```
moritzmeister, line 445: Same as for python part, we don't want to surface this to the user and if so he sets it once when creating the connection object already.

moritzmeister (changes requested):

java/src/main/java/com/logicalclocks/hsfs/TrainingDataset.java
kennethmhc: @moritzmeister In Java client, engine is not specified when creating the connection. Also, hive engine is not available. I think the user need to specify it in this case.

moritzmeister: ah yes I am stupid, there is not Hive engine in scala. so this is actually irrelevant. no need for changes in the Java part of the client. Sorry about that. you can keep the setting in the lower level classes for the future, but simply always pass False, since it's always Spark.

python/hsfs/training_dataset.py
```diff
@@ -633,7 +633,7 @@ def get_query(self, online: bool = True, with_label: bool = False):
             `str`. Query string for the chosen storage used to generate this training
                 dataset.
         """
-        return self._training_dataset_engine.query(self, online, with_label)
+        return self._training_dataset_engine.query(self, online, with_label, engine.get_type())
```
moritzmeister, line 636: Doesn't the `get_type()` return a string with `"hive"` and not a boolean?

kennethmhc: @moritzmeister Absolutely! just fixed that.

moritzmeister (approved):

PR #536, opened 2021-12-01
[HOPSWORKS-2741] Kafka Storage c

moritzmeister:


SirOibaf:
@moritzmeister - you forgot to push the Java/Scala version.

java/src/main/java/com/logicalclocks/hsfs/StorageConnector.java
```diff
@@ -379,4 +384,106 @@ public String getPath(String subPath) {
...
+      return options;
+    }
+
+    @JsonIgnore
+    public String getPath(String subPath) {
+      return null;
+    }
+
+    public Dataset<Row> read(String query, String dataFormat, Map<String, String> options, String path) {
+      throw new NotSupportedException("Reading a Kafka Stream into a static Spark Dataframe is not supported.");
```
kennethmhc, line 465: Why we do not support this? Spark can write kafka source to static DF.
https://spark.apache.org/docs/latest/structured-streaming-kafka-integration.html#creating-a-kafka-source-for-batch-queries

moritzmeister: Mostly cause I thought it will not be useful. Is it common to read a snapshot of a kafka topic, and derive features from it?

kennethmhc: I can think of features for example "count_last_hour", "count_last_5_mins". Can it be derived from streaming df as well?

SirOibaf (approved):

PR #550, opened 2022-01-18
[HOPSWORKS-2933] Improve HSFS Python upload memory usage

moritzmeister:
Removes one of the copies of the data in memory


kennethmhc (approved):
LGTM!

PR #554, opened 2022-01-26
[HOPSWORKS-2948] Rewrite feature store name

kennethmhc:
- `get_feature_store()` should require only the project name - the HSFS API should add `_featurestore` and lowercase the project name.
- we need to make it retrocompatible. As such, we should check if the user already provided the `_featurestore` part, and if so, not add it.

### DIFF
--- a/python/hsfs/connection.py
+++ b/python/hsfs/connection.py
@@ -155,8 +155,7 @@ class Connection:
         """
         if not name:
             name = client.get_instance()._project_name
-        return self._feature_store_api.get(
-            util.rewrite_feature_store_name(name))
+        return self._feature_store_api.get(util.rewrite_feature_store_name(name))
 
     @not_connected
     def connect(self):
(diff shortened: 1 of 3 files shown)


python/hsfs/connection.py
```diff
@@ -153,8 +154,8 @@ def get_feature_store(self, name: str = None):
             `FeatureStore`. A feature store handle object to perform operations on.
```
SirOibaf, line 154: @kennethmhc  - can you please update the documentation making it clear that we just need the project name?

SirOibaf (changes requested):
PR looks ok, please improve the documentation.

python/hsfs/connection.py
kennethmhc: @SirOibaf Changed the documentation.

SirOibaf (approved):

PR #569, opened 2022-02-09
[HOPSWORKS-2971] Time Series Split for training datasets

davitbzh:


python/hsfs/engine/spark.py
```diff
@@ -869,6 +871,22 @@ def _apply_transformation_function(self, training_dataset, dataset):
...
+            "rank",
+            percent_rank().over(
+                window.Window.partitionBy().orderBy(
+                    training_dataset._querydto._left_feature_group.event_time
+                )
+            ),
+        )
+        for split_weight in split_weights:
+            feature_dataframe_list.append(
+                dataset.where("rank <= {:.2f}".format(split_weight)).drop("rank")
```
kennethmhc, line 886: rank should be in range `[accumulate_weight, accumulate_weight+ split_weight]`?

davitbzh: ahh yes. good point

python/hsfs/feature_store.py
```diff
@@ -604,13 +605,15 @@ def create_training_dataset(
...
-            transformation_functions: A dictionary mapping tansformation functions to
+            transformation_functions: A dictionary mapping transformation functions to
                 to the features they should be applied to before writing out the
                 training data and at inference time. Defaults to `{}`, no
                 transformations.
             train_split: If `splits` is set, provide the name of the split that is going
                 to be used for training. The statistics of this split will be used for
                 transformation functions if necessary. Defaults to `None`.
+            split_type: If `splits` is set, provide the type of the method to split datasets. Currently `random` and
+            `time_series` types are supported. Defaults to `random`.
```
kennethmhc, line 616: maybe include a description about "time_series" split, it is not so intuitive until I read the code.

PR #573, opened 2022-02-14
[HOPSWORKS-2994] ModuleNotFoundError when saving code

bubriks:


python/hsfs/core/code_engine.py
```diff
@@ -50,7 +50,10 @@ def save_code(self, metadata_instance):
         # JOB
         job_name = os.environ.get(CodeEngine.JOB_ENV)
         # DATABRICKS
-        databricks = importlib.util.find_spec("pyspark.dbutils")
+        try:
+            databricks = importlib.util.find_spec("pyspark.dbutils")
+        except ModuleNotFoundError:
+            pass
```
SirOibaf, line 56: Replace this with `databricks = false` otherwise you'll get an exception below.

SirOibaf (changes requested):

SirOibaf (approved):

PR #578, opened 2022-02-18
[HOPSWORKS-2918] Remove register builtin functions

moritzmeister:
On Master we are removing the function entirely, on the 2.5 release we will add a deprecation note that it's not needed anymore.﻿


moritzmeister:
> 

I fixed the notebook and removed the section with the registering of the functions. We also have an example showing the use of a regular transformation function which is not built in https://examples.hopsworks.ai/master/featurestore/hsfs/online_transformations/online_transformations/

vatj (approved):

PR #589, opened 2022-03-03
[HOPSWORKS-2879] Amend upgrade hudi to 0.10.0 release

davitbzh:
fix time formatting issue


SirOibaf:
If Hudi now supports milliseconds for commit timestamps, should we remove the `/ 1000` here: https://github.com/logicalclocks/feature-store-api/blob/e2da71ee7bf2e8e3a7dd97057c051127a5de4860/python/hsfs/util.py#L128 and in the Java part?

davitbzh:
> If Hudi now supports milliseconds for commit timestamps, should we remove the `/ 1000` here:
> 
> https://github.com/logicalclocks/feature-store-api/blob/e2da71ee7bf2e8e3a7dd97057c051127a5de4860/python/hsfs/util.py#L128
> 
> and in the Java part?

Updated but not sure why python datetime formatting behaves so strange. It reports back microseconds so I had to strip last 000s. anyway this is correct now

SirOibaf:
@davitbzh  - the Java implementation still formats in seconds, not milliseconds. Can you please change it to be consistent?

davitbzh:
> @davitbzh - the Java implementation still formats in seconds, not milliseconds. Can you please change it to be consistent?

I think this is doing that https://github.com/logicalclocks/feature-store-api/pull/589/files#diff-70c6130a5173144ee086290340f2332e26a09ef11e23798e2fec687cc58143e1R167

SirOibaf:
No - https://github.com/logicalclocks/feature-store-api/blob/e2da71ee7bf2e8e3a7dd97057c051127a5de4860/java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java#L92 - this is still wrong.

davitbzh:
> No -
> 
> https://github.com/logicalclocks/feature-store-api/blob/e2da71ee7bf2e8e3a7dd97057c051127a5de4860/java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java#L92
> 
> - this is still wrong.

@SirOibaf thats true, done

SirOibaf (approved):

PR #591, opened 2022-03-08
Fix td.get_query() for queries that contain hudi/on-demand fgs - Scala

SirOibaf:


java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
```diff
@@ -208,15 +209,20 @@ private FeatureGroupCommit getLastCommitMetadata(SparkSession sparkSession, Stri
...
       hudiArgs.put(HUDI_BEGIN_INSTANTTIME_OPT_KEY, timeStampToHudiFormat(startTimestamp));
     } else {
       hudiArgs.put(HUDI_BEGIN_INSTANTTIME_OPT_KEY, timeStampToHudiFormat(0L));
     }
 
-    hudiArgs.put(HUDI_END_INSTANTTIME_OPT_KEY, timeStampToHudiFormat(endTimestamp));
+    if (endTimestamp != null) {
+      hudiArgs.put(HUDI_END_INSTANTTIME_OPT_KEY, timeStampToHudiFormat(endTimestamp));
+    } else {
+      hudiArgs.put(HUDI_END_INSTANTTIME_OPT_KEY, timeStampToHudiFormat(System.currentTimeMillis()));
```
moritzmeister, line 223: Why did this change?

moritzmeister (approved):
I am approving, but one question

java/src/main/java/com/logicalclocks/hsfs/engine/HudiEngine.java
SirOibaf: When returning the training dataset query, the end timestamp is empty. But Hudi needs an "end" timestamp to decide which commits to include. I think that for batch inference we should include all the commits.

moritzmeister (approved):

PR #593, opened 2022-03-17
Prevent side effect on pandas dataframe column names

### LINKED ISSUE #575: DataFrame Column Names are changed on save and insert
When saving or inserting a pandas DataFrame into a FeatureGroup, the column names are changed to lower case on the DataFrame in-place. This is a bit confusing and will likely break any further processing or operations on the data frame after saving/inserting it into the Feature Store. 

Here is a minimal example:
```python
import hsfs
import pandas as pd

d = {'COL_A': [1, 2], 'COL_B': [3, 4]}
df = pd.DataFrame(data=d)

print(f"columns={df.columns}") # -> columns=Index(['COL_A', 'COL_B'], dtype='object')

connection = hsfs.connection()
fs = connection.get_feature_store()
fg = fs.get_feature_group(name=fg_name, version=fg_version)
fg.insert(df)  # same for fg.save(df)

print(f"columns={df.columns}") # -> columns=Index(['col_a', 'col_b'], dtype='object')
```
If I try to access columns the same way as before calling `save`, the code breaks. This is an unexpected side effect.
If it is required for the Feature Store to convert all column names to lower case, then this should be done on a copy of the DataFrame rather than the original object.

moritzmeister:
closes #575

### DIFF
--- a/python/hsfs/engine/hive.py
+++ b/python/hsfs/engine/hive.py
@@ -18,6 +18,8 @@ import pandas as pd
 import numpy as np
 import boto3
 import time
+import re
+import warnings
 
 from io import BytesIO
 from pyhive import hive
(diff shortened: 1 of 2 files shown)


kennethmhc (approved):
LGTM!

PR #595, opened 2022-03-18
[HOPSWORKS-2999] HSFS Hive engine should throw proper errors for all unsupported methods

### LINKED ISSUE #280: Creating TD in Python env from a pandas dataframe should throw NotSupportedException
Currently we only support creating training datasets from Query objects in the Python environment. When a user tries to create it form a dataframe instead she ends up with this AttributeError:
image

This is not very informative and instead should throw an error telling the user to use the Query API.

moritzmeister:
closes #280


kennethmhc (approved):
LGTM!

PR #601, opened 2022-03-31
[HOPSWORKS-2543] AMEND:  return Object types from StorageConnector read

davitbzh:


DhananjayMukhedkar (commented):
Review and tested for `StorageConnector.java`

moritzmeister (approved):

PR #603, opened 2022-04-03
Switch to Parquet when ingesting data from Python

SirOibaf:


moritzmeister (approved):
Look ok in general, one thing - I wonder if we still need all the schema handling in the helper job.

PR #617, opened 2022-04-12
Hsfs generics

davitbzh:


SirOibaf:
Is there a design document of what you want to achieve with this PR?

PR #624, opened 2022-04-20
update transformation function docs

davitbzh:


docs/templates/transformation_functions.md
```diff
@@ -11,39 +12,29 @@ or [Hopsworks job](https://hopsworks.readthedocs.io/en/stable/user_guide/hopswor
...
-
-    !!! example "Register built-in transformation functions in the Hopsworks feature store."
-        ```python
-        fs.register_builtin_transformation_functions()
-        ```
-
 To retrieve all transformation functions from the feature store use `get_transformation_functions` which will return list of `TransformationFunction` objects.
 A specific transformation function can be retrieved by `get_transformation_function` method where the user can provide a name and a version of the transformation function.
-If only the name is provided then it will default to version 1.
+If only function name is provided then it will default to version 1.
```
jimdowling, line 37: only the function name is provided

jimdowling (approved):

PR #627, opened 2022-04-22
Set Hudi default parallelism options if user didn't specify them

davitbzh:


java/src/main/java/com/logicalclocks/hsfs/engine/hudi/HudiEngine.java
```diff
@@ -134,6 +142,12 @@ public void saveHudiFeatureGroup(SparkSession sparkSession, FeatureGroupBase fea
 
     Map<String, String> hudiArgs = setupHudiWriteOpts(featureGroup, operation, writeOptions);
 
+    for (Map.Entry<String, String> entry : HUDI_DEFAULT_PARALLELISM.entrySet()) {
+      if (writeOptions != null && !writeOptions.containsKey((entry.getKey()))) {
+        hudiArgs.put(entry.getKey(), entry.getValue());
+      }
+    }
```
SirOibaf, line 149: We have a method called `setHudiWriteOptions` why don't we set those options there? Sot that they are reused across (e.g. by the delete method as well) and not just by the save method.

Also as for the other Hudi options, you can just set them and let the user overwrite them with the `putAll` call we have here - https://github.com/logicalclocks/feature-store-api/blob/1a421731b0c45af0e4f1a02e96ecfcb93fa10488/java/src/main/java/com/logicalclocks/hsfs/engine/hudi/HudiEngine.java#L249

python/hsfs/core/hudi_engine.py
```diff
@@ -112,6 +117,12 @@ def __init__(
     def save_hudi_fg(
         self, dataset, save_mode, operation, write_options, validation_id=None
     ):
+
+        for parallelism_option in HudiEngine.HUDI_DEFAULT_PARALLELISM:
+            if parallelism_option not in write_options:
+                write_options[parallelism_option] = HudiEngine.HUDI_DEFAULT_PARALLELISM[
+                    parallelism_option
+                ]
```
SirOibaf, line 125: Same comment as for the Java code.

SirOibaf (changes requested):

java/src/main/java/com/logicalclocks/hsfs/engine/hudi/HudiEngine.java
davitbzh: regarding `hudiArgs.putAll(writeOptions)` my idea was if they provide some option we don't override them.

SirOibaf: Yes you are correct that we should let the user overwrite the options. My argument is that you can just add the following code
```
 for (Map.Entry<String, String> entry : HUDI_DEFAULT_PARALLELISM.entrySet()) {
          hudiArgs.put(entry.getKey(), entry.getValue());
      }
```
and then if the user wants to overwrite them, the `putAll` will take care of setting the user preference, but I believe what I'm suggesting is a cleaner way of writing the code.

SirOibaf (approved):

PR #632, opened 2022-04-26
Improve features documentation

SirOibaf:


docs/templates/feature.md
```diff
@@ -2,42 +2,41 @@
...
-New features can be appended to [feature groups](feature_group.md), however, to drop
-features, a new [feature group](feature_group.md) version has to be created. When
-appending features it is possible to specify a default value which is used for existing
-feature vectors in the [feature group](feature_group.md) for the new feature.
+## Features Taxonomy
 
-{{feature}}
+Within a [feature group](feature_group.md) there are different categories of features:
+
+* **Primary Keys**: These features describe the entity contained in a feature group (e.g. customer features, user features, ...). Within a feature group there could be more than one feature describing the primary key.
```
jimdowling, line 15: A primary key is not a feature. It is not used to train a model. It is only used for online feature store for retrieving features. I prefer to use the term 'column' for that reason.

The columns that uniquely identify an entity in a feature group (e.g., the customer, the transaction, ...). A primary key may consist of multiple columns, that is, a composite primary key.

docs/templates/feature.md
```diff
@@ -2,42 +2,41 @@
...
+* **Event Time**: One feature within a feature group can be selected to identify the time at which the event for the given record has happened. The event time feature is going to be used to enforce Point-in-time correctness when joining features from different feature groups together. When storing feature data on the offline feature store, each record is uniquely identified by the union of the primary key features and the even time feature.
```
jimdowling, line 16: event time is not a feature either. it is not used to train models. You don't select it - you identify it. Select is an overloaded term, best to avoid when talking about databases.

The column in a feature group that stores the time at which the event for the given record has happened (the observation time for the row). You need to provide the event time, if you want the feature group to be able to support point-in-time correct joins when used to creating training data.

docs/templates/feature.md
```diff
@@ -2,42 +2,41 @@
...
+* **Partition Keys**: These features describe the storage layout of the feature group data on the offline feature store. Partition keys allow to divide the feature group data into different subdirectory. Partitioning can help to query more efficiently the data from the offline feature store.
```
jimdowling, line 17: the columns that define the storage layout of the feature group data in the offline feature store. The partition keys may be defined over one or more columns, and the storage layout is based on [hive-style partitioning](https://athena.guide/articles/hive-style-partitioning/), that is, partitions are stored in separate subdirectories. Partitioning can help improve query efficiency for the offline feature store, by reducing the amount of data that needs to be read from disk for a given query.

docs/templates/feature.md
```diff
@@ -2,42 +2,41 @@
...
+* **Normal Features**: These features represent attributes of related to the primary keys.
```
jimdowling, line 18: ** Normal Features ** -> ** Features **
These columns are the features that can be used to train models and perform inference on models.

docs/templates/feature.md
```diff
@@ -2,42 +2,41 @@
...
 
 ## Feature Types
```
jimdowling, line 20: These are not feature types.
These are "Data Storage Types" in the online and offline stores.
A feature type is - categorical, numerical, embedding, etc.

docs/templates/feature.md
```diff
@@ -2,42 +2,41 @@
...
 
-Each features requires at least an offline type to be specified for the creation of the
-meta data of the [feature group](feature_group.md) in the offline storage, even if the
-[feature group](feature_group.md) is going to be a purely online
-[feature group](feature_group.md) with no data in the offline storage.
+Each features has two different data types:
```
jimdowling, line 22: When a feature is stored in the both the online and offline feature stores, it will be stored in a data storage type native to each store.

docs/templates/feature.md
```diff
@@ -2,42 +2,41 @@
...
+
+* **Offline type**: The data type of the feature when stored on the offline feature store
```
jimdowling, line 24: Offline data type

docs/templates/feature.md
```diff
@@ -2,42 +2,41 @@
...
+* **Online type**: The data type of the feature when stored on the online feature store.
```
jimdowling, line 25: Online data type

docs/templates/feature.md
```diff
@@ -2,42 +2,41 @@
...
+
+The offline type is always required, even if the feature group is stored only online. On the other hand, if the feature group is not *online_enabled*, its features will not have an online type.
```
jimdowling, line 27: online data type
offline data type

docs/templates/feature.md
```diff
@@ -51,27 +50,77 @@ Potential *offline* types are:
 
 ### Online Storage
 
-The online storage is based on MySQL Cluster (NDB) and hence, any
-[MySQL Data Type](https://dev.mysql.com/doc/refman/8.0/en/data-types.html)
+The online storage is based on RonDB and hence, any
+[RonDB Data Type](https://dev.mysql.com/doc/refman/8.0/en/data-types.html)
```
jimdowling, line 54: MySQL Data Type

docs/templates/feature.md
```diff
@@ -51,27 +50,77 @@ Potential *offline* types are:
...
-"CHAR", "VARCHAR(25)", "VARCHAR(125)", "VARCHAR(225)", "VARCHAR(500)",
-"VARCHAR(1000)", "VARCHAR(2000)", "VARCHAR(5000)", "VARCHAR(10000)", "BINARY",
-"VARBINARY(100)", "VARBINARY(500)", "VARBINARY(1000)", "BLOB", "TEXT",
-"TINYBLOB", "TINYTEXT", "MEDIUMBLOB", "MEDIUMTEXT", "LONGBLOB", "LONGTEXT",
-"JSON"
+"CHAR", "VARCHAR(n)", "BINARY", "VARBINARY(n)", "BLOB", "TEXT", "TINYBLOB",
+"TINYTEXT", "MEDIUMBLOB", "MEDIUMTEXT", "LONGBLOB", "LONGTEXT", "JSON"
 ```
 
+#### Complex online types
```
jimdowling, line 66: online data types

docs/templates/feature.md
```diff
@@ -2,42 +2,41 @@
...
-[feature group](feature_group.md) is going to be a purely online
-[feature group](feature_group.md) with no data in the offline storage.
+Each features has two different data types:
+
+* **Offline type**: The data type of the feature when stored on the offline feature store
+* **Online type**: The data type of the feature when stored on the online feature store.
+
+The offline type is always required, even if the feature group is stored only online. On the other hand, if the feature group is not *online_enabled*, its features will not have an online type.
 
 ### Offline Storage
```
jimdowling, line 29: Offline Primitive Data Types

docs/templates/feature.md
```diff
@@ -51,27 +50,77 @@ Potential *offline* types are:
 
 ### Online Storage
```
jimdowling, line 51: Online Primitive Data Types

docs/templates/feature.md
```diff
@@ -51,27 +50,77 @@ Potential *offline* types are:
...
-"TINYBLOB", "TINYTEXT", "MEDIUMBLOB", "MEDIUMTEXT", "LONGBLOB", "LONGTEXT",
-"JSON"
+"CHAR", "VARCHAR(n)", "BINARY", "VARBINARY(n)", "BLOB", "TEXT", "TINYBLOB",
+"TINYTEXT", "MEDIUMBLOB", "MEDIUMTEXT", "LONGBLOB", "LONGTEXT", "JSON"
 ```
 
+#### Complex online types
+
+Additionally to the *online* types above, Hopsworks allows users to store complex types (e.g. *ARRAY<INT>*) on the online feature store.
+Hopsworks will take care of transparently serializing the complex features and storing them as *VARBINARY* on the online feature store. The serialization happens when calling the [save()](../api/feature_group_api/#save), [insert()](../api/feature_group_api/#insert) or [insert_stream()](../api/feature_group_api/#insert_stream) methods. The deserialization will be executed when calling the [get_serving_vector()](../api/training_dataset_api/#get_serving_vector) method to retrieve data from the online feature store.
```
jimdowling, line 69: Hopsworks will take care of transparently serializing  ->
Hopsworks serializes the complex features transparently  and stores them as *VARBINARY* in the online feature store.

docs/templates/feature.md
```diff
@@ -51,27 +50,77 @@ Potential *offline* types are:
...
+If users query directly the online feature store, for instance using the `fs.sql("SELECT ...", online=True)` statement, they will receive back a binary blob.
```
jimdowling, line 70: they will receive back a binary blob ->
it will return a binary blob

docs/templates/feature.md
```diff
@@ -51,27 +50,77 @@ Potential *offline* types are:
...
+
+On the feature store UI, the online feature type for complex features will be reported as *VARBINARY*.
+
+#### Online restrictions for primary key types:
+
+When a feature is being used as a primary key, certain types are not allowed. Examples of such types are *Float*, *Double*, *Date*, *Text*, *Blob* and *Complex Types*  (e.g. Array<>). Additionally the size of the sum of the primary key online types storage requirements should not exceed 3KB.
```
jimdowling, line 76: Some primitive data types are not supported as a part of the primary key. These are:  *Float*, *Double*, *Date*, *Text*, *Blob* and *Complex Types*  (e.g. Array<>). Additionally the size of the sum of the primary key online types storage requirements should not exceed 3KB.

Don't use examples. State explicitly which ones are supported. 
In this case, it might be easier to say which primitive data types are supported - string, int, datetime, etc.

docs/templates/feature.md
```diff
@@ -51,27 +50,77 @@ Potential *offline* types are:
...
+
+### Type Inference
+
+The offline and online types for each feature are inferred automatically when calling the [save()](../api/feature_group_api/#save) method to create a feature group. The types will be inferred based on the types of the Spark or Pandas DataFrame.
```
jimdowling, line 80: The offline and online types for each column are identified based on the types of the columns in the Spark or Pandas DataFrame, and those types are then mapped to the online and offline data types.

docs/templates/feature.md
```diff
@@ -51,27 +50,77 @@ Potential *offline* types are:
...
+Hopsworks will take care of transparently serializing the complex features and storing them as *VARBINARY* on the online feature store. The serialization happens when calling the [save()](../api/feature_group_api/#save), [insert()](../api/feature_group_api/#insert) or [insert_stream()](../api/feature_group_api/#insert_stream) methods. The deserialization will be executed when calling the [get_serving_vector()](../api/training_dataset_api/#get_serving_vector) method to retrieve data from the online feature store.
+If users query directly the online feature store, for instance using the `fs.sql("SELECT ...", online=True)` statement, they will receive back a binary blob.
+
+On the feature store UI, the online feature type for complex features will be reported as *VARBINARY*.
+
+#### Online restrictions for primary key types:
+
+When a feature is being used as a primary key, certain types are not allowed. Examples of such types are *Float*, *Double*, *Date*, *Text*, *Blob* and *Complex Types*  (e.g. Array<>). Additionally the size of the sum of the primary key online types storage requirements should not exceed 3KB.
+
+### Type Inference
```
jimdowling, line 78: Technically, this is not Type Inference. It is Type Mapping. We are not inferring the type of columns, we have a mapping from a Pandas/Spark type to an online/offline data type.
I think Type Inference as a subtitle is helpful for programmers to understand what is going on, but be careful not to say we infer types. We don't, we map types.

docs/templates/feature.md
```diff
@@ -51,27 +50,77 @@ Potential *offline* types are:
...
+
+The offline and online types for each feature are inferred automatically when calling the [save()](../api/feature_group_api/#save) method to create a feature group. The types will be inferred based on the types of the Spark or Pandas DataFrame.
+
+In the case of Spark DataFrame, the [Spark types](https://spark.apache.org/docs/latest/sql-ref-datatypes.html) will be converted in the equivalent Hive Metastore type and used as offline feature store type. If the feature group is online enabled, Hopsworks will then match the offline type to the corresponding online type. The matching is performed as following:
```
jimdowling, line 82: of a Spark DataFrame
converted in -> mapped to the corresponding 
used as offline feature store type -> used as the offline data type

will then match the offline type to the corresponding online type. The matching is performed as following:
->
will then map the offline data type to the corresponding online data type. The mapping is based on the following rules:

docs/templates/feature.md
```diff
@@ -51,27 +50,77 @@ Potential *offline* types are:
...
+
+* If the offline type is supported also on the online feature store (e.g. INT, FLOAT, DATE, TIMESTAMP), the online type will be equivalent to the offline type
```
jimdowling, line 84: supported also -> also supported

jimdowling: be equivalent to -> the same as

docs/templates/feature.md
```diff
@@ -51,27 +50,77 @@ Potential *offline* types are:
...
+* If the offline type is *boolean*, the online type is going to be set as *tinyint*
```
jimdowling, line 85: change all "online type" and "offline type" to "online data type" and "offline data type"

docs/templates/feature.md
```diff
@@ -51,27 +50,77 @@ Potential *offline* types are:
...
+* If the offline type is *string*, the online type is going to be set as *varchar(100)*
+* If the offline type is not supported by the online feature store and it is not one of the above exception, the online type will be set as *varbinary(100)* to handle complex types.
+
+#### Pandas Conversion
```
jimdowling, line 89: Conversion is ok here.

docs/templates/feature.md
```diff
@@ -51,27 +50,77 @@ Potential *offline* types are:
...
+
+When registering a [Pandas](https://pandas.pydata.org/) DataFrame as feature group, the following conversion is applied:
```
jimdowling, line 91: as a feature group
conversion -> mapping rules are applied

docs/templates/feature.md
```diff
@@ -51,27 +50,77 @@ Potential *offline* types are:
...
+| ------------------ | ------------------- |
+| int32              | INT                 |
+| int64              | BIGINT              |
+| float32            | FLOAT               |
+| float64            | DOUBLE              |
+| datetime64[ns]     | TIMESTAMP           |
+| object             | STRING              |
+
+
+### Type Overwrite
```
jimdowling, line 103: Explicit schema definition

docs/templates/feature.md
```diff
@@ -51,27 +50,77 @@ Potential *offline* types are:
...
+
+When creating a feature group it is possible for the user to control both the offline and online type of each feature. If users manually specify a schema for the feature group, Hopsworks is going to use it to create the feature group, without performing any type inference.
```
jimdowling, line 105: feature -> column
manually -> explicitly define the schema
use it -> use that schema

type inference is ok here, i think. You could also say - without automatically mapping the user defined schema to the offline and online data types

docs/templates/feature.md
```diff
@@ -51,27 +50,77 @@ Potential *offline* types are:
...
+Users can manually define the feature group schema as follow:
```
jimdowling, line 106: explicitly

jimdowling: as follows

jimdowling (changes requested):

docs/templates/feature.md
SirOibaf: I changed it, though I don't fully agree. While technically correct that you don't use primary keys as features to train the model, introducing the word "column" in the documentation I think just creates confusion for new users - those reading the documentation in the first place. 
We also don't have different APIs to define what's a feature and what's a primary key, same thing in the UI, everything is referred as feature - mostly for simplicity and ease of use.

So in a sense, by changing the word "feature" with the word "column" we are trading simplicity for correctness.

docs/templates/feature.md
SirOibaf: hive-style partitioning is technically incorrect. Hudi doesn't use hive style partitioning by default, it doesn't use `key=value`, but rather just `value`. Hudi does support hive style partitioning, but you have to enable it explicitly.

docs/templates/feature.md
jimdowling: Primary keys, event_time, and features are all columns. 
Maybe we need to define that earlier in the docs. We have to differentiate them - it's not simpler to confuse them. I agree we should minimize using the word column, but if we have to refer to the collection of Primary keys, event_time, and features - i don't see what other word to use.

I would also note another important terminology issue. The primary key for RonDB is not the same as Hudi. In RonDB, it doesn't include event_time, but it does in Hudi. For this reason, it the book, i refer to the PK as the "entity key(s)". I call it a primary key when we get to HSFS. See more in the book, see section 1.1.1.

docs/templates/feature.md
jimdowling: ok

jimdowling (approved):

PR #641, opened 2022-05-19
[HOPSWORKS-3159] apply transformations in python kernel

davitbzh:


python/hsfs/core/statistics_engine.py
```diff
@@ -96,10 +98,15 @@ def profile_transformation_fn_statistics(feature_dataframe, columns):
                 "statistics for. A possible cause might be that you inserted only data "
                 "to the online storage of a feature group."
             )
-        return engine.get_instance().profile(
+        content_str = engine.get_instance().profile(
             feature_dataframe, columns, False, True, False
         )
 
+        # add unique value profile to String type columns
+        return self.profile_unique_values(
```
kennethmhc, line 106: move this to `python.profile` instead, because this only works in python engine.

kennethmhc: If you call it directly from the class like `StatisticsEngine.profile_unique_values`, you can keep the method static.

python/hsfs/core/builtin_transformation_function.py
```diff
@@ -80,8 +80,11 @@ def robust_scaler_stats(content, feature_name):
...
     @staticmethod
     def encoder_stats(content, feature_name):
+        value_to_index = {}
         for col in content["columns"]:
-            if col["column"] == feature_name:
-                unique_data = [i["value"] for i in col["histogram"]]
-                value_to_index = dict((c, i) for i, c in enumerate(unique_data))
+            if col["column"] == feature_name and "unique_values" in col:
+                unique_data = [value for value in col["unique_values"]]
+                value_to_index = dict(
```
kennethmhc, line 87: can `return` here?

kennethmhc (changes requested):

python/hsfs/core/statistics_engine.py
davitbzh: it works also for spark engine. Spark engine also returns array of unique values with the mothod `get_unique_values`, similar what python is doing and then it assembles deequ style json

kennethmhc: In spark engine we are using deequ to get unique values, right? I mean we should be able to get unique value in python engine by calling 
```
engine.get_instance().profile(
            feature_dataframe, columns, False, True, True
        )
```

tdoehmen: Since I just saw the conversation here, I can maybe help to clarify. Currently, the deequ counts the number of unique values, but does not capture the values themselves.

davitbzh: @tdoehmen thanks for clarification, thats was also my understanding. @kennethmhc in sprark engine I do the following to get unique values:
```
    def get_unique_values(feature_dataframe, feature_name):
        return feature_dataframe.select(feature_name).distinct().collect()
 ```

Comment (kennethmhc) [reply python/hsfs/core/statistics_engine.py:106]: Got it! Is the unique value being used in `encoder` transformation? `encoder` is using "histogram" right now, need to change it?

File: python/hsfs/engine/python.py [hunk 3]
```diff
@@ -466,7 +473,15 @@ def _split_df(self, df, splits):
         random.shuffle(groups)
         df[split_column] = groups
         for i, item in enumerate(items):
-            result_dfs[item[0]] = df[df[split_column] == i].drop(split_column, axis=1)
+            split_df = df[df[split_column] == i].drop(split_column, axis=1)
+            if item == training_dataset_obj.train_split:
```
kennethmhc, line 477: May be move the transformation part out of this method? Now this method not only do "_split_df".

davitbzh: Not sure if I follow: here  `populate_builtin_transformation_functions` needs to know which split to use and `_apply_transformation_function` is applied one by one to splits which is different if there are no splits

kennethmhc: I mean keep this `split_df` method unchange but `populate_builtin_transformation_functions` and `_apply_transformation_function` in `get_training_data`.

davitbzh: well I need to get `train_split`. under `if training_dataset_obj.splits` I just know that there is split and then I have to extract `train_split` stats. thats why I put this under a loop in `_split_df` to get that info .

kennethmhc: I think the code is almost the same, the difference is just where to put it. I think it is clearer to put it outside of this method because reader may think that transformation is not applied when there is a split.. But I am ok with current implementation as well, but please rename this method name so that it is clear that it also apply transformation.

davitbzh: done

python/hsfs/feature_view.py
```diff
@@ -105,6 +105,21 @@ def init_serving(
...
+        """Initialise and cache parametrized transformation functions.
+
+        # Arguments
+            training_dataset_version: int, optional. Default to be 1. Transformation statistics
+                are fetched from training dataset and apply in serving vector.
+        """
+        self._vector_server = vector_server.VectorServer(
+            self._featurestore_id, training_dataset_version
+        )
+        self._vector_server.init_batch_scoring(self)
```
kennethmhc, line 121: You may need 2 `_vector_server` instance, because `get_feature_vector` won't work if users call `init_batch_scoring` first then call `get_feature_vector`

python/hsfs/core/feature_view_engine.py
```diff
@@ -373,12 +373,28 @@ def delete_training_dataset_only(
                 feature_view_obj.name, feature_view_obj.version
             )
 
-    def get_batch_data(self, feature_view_obj, start_time, end_time, read_options=None):
+    def get_batch_data(
+        self,
+        feature_view_obj,
+        start_time,
+        end_time,
+        vector_server,
```
kennethmhc, line 381: can be `training_dataset_version` instead of `vector_server`

python/hsfs/feature_view.py
davitbzh: yes thats, but its similar what happens if users call  `init_serving(batch=True)` and then they call `get_feature_vector`. We need explain this well in the docs

python/hsfs/feature_view.py
```diff
@@ -100,10 +102,16 @@ def init_serving(
                 If set to False, the online feature store storage connector is used
                 which relies on the private IP.
         """
-        self._vector_server = vector_server.VectorServer(
-            self._featurestore_id, training_dataset_version
-        )
-        self._vector_server.init_serving(self, batch, external)
+        if not batch:
```
kennethmhc, line 105: `batch` can be removed? We should always initiate both instances.

davitbzh: aha OK so you mean when ever `init_serving` is called we initiate both? makes sense

python/hsfs/feature_view.py
```diff
@@ -100,10 +102,32 @@ def init_serving(
                 If set to False, the online feature store storage connector is used
                 which relies on the private IP.
         """
-        self._vector_server = vector_server.VectorServer(
+        # initiate single vector server
```
kennethmhc, line 105: `batch` in the method parameter is not needed.

python/hsfs/engine/python.py
```diff
@@ -479,7 +488,15 @@ def _split_df(self, df, splits):
...
+                transformation_function_engine.TransformationFunctionEngine.populate_builtin_transformation_functions(
+                    training_dataset_obj, feature_view_obj, split_df
+                )
+            result_dfs[item[0]] = self._apply_transformation_function(
```
kennethmhc, line 496: If first `item[0]` is not `training_dataset_obj.train_split`, this does not work.

python/hsfs/engine/python.py
```diff
@@ -479,7 +488,15 @@ def _split_df(self, df, splits):
         random.shuffle(groups)
         df[split_column] = groups
         for i, item in enumerate(items):
-            result_dfs[item[0]] = df[df[split_column] == i].drop(split_column, axis=1)
+            split_df = df[df[split_column] == i].drop(split_column, axis=1)
+            if item == training_dataset_obj.train_split:
```
kennethmhc, line 492:
```suggestion
            if item[0] == training_dataset_obj.train_split:
```

kennethmhc (dismissed):

python/hsfs/engine/python.py
davitbzh: good catch

python/hsfs/engine/python.py
kennethmhc: I mean first `item` may not guarantee to be `train_split`?

davitbzh: `item[0]` is just a tuple with split name and ratio. here we are just constructing dict  `result_dfs` which is split name and dataframe. Not sure why it has to be  `train_split`?

jimdowling (approved):

PR #643, opened 2022-05-23
[HOPSWORKS-3163] Feature View API: return td split dataframes and infer csv schema in spark

davitbzh:


python/hsfs/engine/spark.py
```diff
@@ -448,6 +448,13 @@ def read(self, storage_connector, data_format, read_options, location):
 
             if data_format.lower() == "tsv":
                 data_format = "csv"
+
+            if data_format.lower() == "csv":
```
kennethmhc, line 452: maybe we maybe add default write and read option in each read and write methods in feature_view_engine, similar to https://github.com/logicalclocks/feature-store-api/blob/master/python/hsfs/core/training_dataset_engine.py#L102

kennethmhc (approved):
LGTM!

PR #651, opened 2022-05-31
Fix update description for stream feature groups

SirOibaf:


moritzmeister (approved):

moritzmeister (changes requested):
Don't we need to do the same in `update_description` then? we also make a copy of the object there

moritzmeister (approved):

PR #654, opened 2022-06-02
[HOPSWORKS-3099] Feature view Training dataset on Java client

kennethmhc:
Changes:
1. Implement training dataset operation in feature view
 - get, create, delete, recreate
2. Implement taining_dataset_bundle
3. Calculate statistics of feature view training dataset


kennethmhc:
example notebook: https://gist.github.com/kennethmhc/2d2f58f5f14a5d4ec7d839df93f59256

java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
```diff
@@ -249,67 +261,149 @@ public void deleteTag(String name) throws FeatureStoreException, IOException {
...
+            .eventStartTime(startTime)
+            .eventEndTime(endTime)
+            .dataFormat(dataFormat)
+            .storageConnector(storageConnector)
+            .build();
+    return featureViewEngine.createTrainingDataset(this, trainingDataset, Maps.newHashMap());
   }
 
   public void createTrainingDataset(
       String startTime, String endTime, DataFormat dataFormat, StorageConnector storageConnector,
```
tdoehmen, line 344: When calling this function fvUpdate.createTrainingDataset(null, null, CSV, null, splits, "train") it returns NULL instead of a TrainingDatasetBundle. I'm not sure why that is. But the training data gets successfully written to the HopsFS storage. Can you have a look at this?

Something a little unrelated, but I wonder why the training data is not coalesced before it's written. The resulting dataset is split into literally one CSV file per row.

tdoehmen (changes requested):

java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
kennethmhc: Good catch! `createTrainingDataset` should return `TrainingDatasetBundle` instead of `void`.
T
`coalesce` is optional and default to be false.

kennethmhc:
@tdoehmen addressed your comments.

tdoehmen (approved):

PR #656, opened 2022-06-03
[HOPSWORKS-3103][HOPSWORKS-3161] Improve hive error message

kennethmhc:


java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java
```diff
@@ -143,7 +142,20 @@ public String getCertKey() {
...
+    try {
+      return sparkSession.sql(query);
+    } catch (Exception e) {
+      if (e.getMessage().contains("Permission denied")) {
+        Pattern pattern = Pattern.compile("inode=\"/apps/hive/warehouse/(.*)?_featurestore\\.db\"");
+        Matcher matcher = pattern.matcher(e.getMessage());
+        if (matcher.find()) {
+          String featureStore = matcher.group(1);
+          throw new RuntimeException(String.format("Cannot access feature store '%s'. " +
+              "It is possible to request access from admins of '%s'.", featureStore, featureStore));
```
SirOibaf, line 154: We don't use "project admins" as a concept, we use "data owner". Let's be consistent here so that users don't get confused.

python/hsfs/engine/python.py
```diff
@@ -532,16 +534,28 @@ def write_training_dataset(
...
+        except (TTransportException, AttributeError):
+            raise ValueError(
+                f"Cannot connect to hive server. Please check the host name '{client.get_instance()._host}' "
+                "is correct and make sure port '9085' is open on host server."
+            )
+        except OperationalError as e:
+            if e.args[0].status.statusCode == 3:
+                raise RuntimeError(
+                    f"Cannot access feature store '{feature_store}'. Please check if your project has the access right."
+                    f" It is possible to request access from admins of '{feature_store}'."
```
SirOibaf, line 557: Same "project admin" comment as above.

SirOibaf (changes requested):

java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java
```diff
@@ -143,7 +142,20 @@ public String getCertKey() {
...
 
   public Dataset<Row> sql(String query) {
-    return sparkSession.sql(query);
+    try {
+      return sparkSession.sql(query);
+    } catch (Exception e) {
+      if (e.getMessage().contains("Permission denied")) {
+        Pattern pattern = Pattern.compile("inode=\"/apps/hive/warehouse/(.*)?_featurestore\\.db\"");
+        Matcher matcher = pattern.matcher(e.getMessage());
+        if (matcher.find()) {
```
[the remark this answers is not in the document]
kennethmhc: I think it is good to have because if the name/format of hive db changes or reasons, `matcher.group(1)` will throw exception.

SirOibaf (approved):

PR #658, opened 2022-06-03
[HOPSWORKS-3135] explicit provenance - feature groups

o-alex:
https://hopsworks.atlassian.net/browse/HOPSWORKS-3135


SirOibaf:
@o-alex - The Java part is missing.

python/hsfs/feature_group.py
```diff
@@ -216,6 +216,68 @@ def get_tags(self):
         """
         return self._feature_group_engine.get_tags(self)
 
+    def get_parent_feature_groups(self):
+        """Get the parent feature groups that were used to generate this feature
+        group. This is based on explicit provenance.
+        """
```
SirOibaf, line 222: The API documentation is based on these docstrings. Update it so that it follows the correct standard, including the documentation on the returned type.

python/hsfs/feature_group.py
```diff
@@ -216,6 +216,68 @@ def get_tags(self):
...
+                    "accessible": link["accessible"],
+                    "deleted": link["deleted"],
+                }
+            )
+        return parent_links
+
+    def get_derived_feature_groups(self):
+        """Get the derived feature groups that used this feature group.
+        This is based on explicit provenance.
+        """
```
SirOibaf, line 241: Same docstring comment

python/hsfs/feature_group.py
```diff
@@ -216,6 +216,68 @@ def get_tags(self):
...
+                        "accessible": link["accessible"],
+                        "deleted": link["deleted"],
+                    }
+                )
+        return derived_feature_groups
+
+    def get_derived_feature_views(self):
+        """Get the derived feature views that used this feature group.
+        This is based on explicit provenance.
+        """
```
SirOibaf, line 264: Same docstring comment

python/hsfs/feature_group.py
```diff
@@ -216,6 +216,68 @@ def get_tags(self):
...
+        group. This is based on explicit provenance.
+        """
+        links_dto = self._feature_group_engine.get_explicit_provenance_links(self)
+        parent_links = []
+        for link in links_dto.upstream:
+            parent_links.append(
+                {
+                    "project": link["project"],
+                    "name": link["name"],
+                    "version": link["version"],
```
SirOibaf, line 230: IMHO this is not a user friendly API. 

- I would return an object and not a dictionary. If you return an object here, the documentation will be able to pick it up and users will be able to know what this method returns, without having to go read the source code.

- I would not call the field project, I would call it `feature_store` to be compatible with the `get_feature_store()` call. and make the API consistent. The project is only using during the connection setup, everywhere else we use the `feature_store` reference.

- If the feature group exists and it's accessible, I would make the call invoke the `get_feature_group` method (from the correct feature store) and store the feature group reference in the object. This will make the APIs more user friendly and consistent with the parent configuration when you create a feature group. When you create a feature group you provide a metadata object, not name/version. So IMHO the method should return the metadata object as well (if possible).

python/hsfs/feature_group.py
```diff
@@ -216,6 +216,68 @@ def get_tags(self):
...
+    def get_derived_feature_groups(self):
+        """Get the derived feature groups that used this feature group.
+        This is based on explicit provenance.
+        """
+        links_dto = self._feature_group_engine.get_derived_artifacts(self)
+        derived_feature_groups = []
+        for link in links_dto.upstream:
+            if (
+                link["type"] == "CACHED_FEATURE_GROUP"
+                or link["type"] == "FEATURE_GROUP"
```
SirOibaf, line 247: What's the difference between these 2? Should it be `STREAM_FEATURE_GROUP`? What about the external feature groups?

python/hsfs/feature_group.py
```diff
@@ -216,6 +216,68 @@ def get_tags(self):
...
+                derived_feature_groups.append(
+                    {
+                        "project": link["project"],
+                        "name": link["name"],
+                        "version": link["version"],
+                        "type": link["type"],
+                        "accessible": link["accessible"],
+                        "deleted": link["deleted"],
+                    }
+                )
```
SirOibaf, line 258: Same comment on the return type as for the method above.

python/hsfs/feature_group.py
```diff
@@ -216,6 +216,68 @@ def get_tags(self):
...
+                derived_feature_views.append(
+                    {
+                        "project": link["project"],
+                        "name": link["name"],
+                        "version": link["version"],
+                        "type": link["type"],
+                        "accessible": link["accessible"],
+                        "deleted": link["deleted"],
+                    }
+                )
```
SirOibaf, line 278: Same comment on the return type as for the method above.

SirOibaf (changes requested):
Reviewed the Python part as the Java part is missing.

Additionally I would add the "get_parent_feature_groups()` in the feature view as well.

python/hsfs/feature_group.py
o-alex: typo, it was indeed STREAM_FEATURE_GROUP.

moritzmeister (approved):

SirOibaf (approved):

PR #664, opened 2022-06-10
[HOPSWORKS-3186] fv td API redesign

kennethmhc:
Changes (only python client, java client in a separate PR):
1. Implement the new fv td API
2. Add time series split to API but not implemented yet, so will throw error if users try to use time series split.
3. Introduce a new class for td split.


kennethmhc:
documentation is not still ongoing.

python/hsfs/feature_view.py
```diff
@@ -418,46 +415,583 @@ def create_training_dataset(
...
+            splits=[TrainingDatasetSplit.TRAIN, TrainingDatasetSplit.TEST],
+        )
+        warnings.warn(
+            "Incremented version to `{}`.".format(td.version),
+            util.VersionWarning,
+        )
+        return df[TrainingDatasetSplit.TRAIN] + df[TrainingDatasetSplit.TEST]
+
+    @staticmethod
+    def _validate_train_test_splits(test_size, train_end, test_start):
```
jimdowling, line 762: splits -> split

python/hsfs/feature_view.py
```diff
@@ -418,46 +415,583 @@ def create_training_dataset(
...
+            (val_size and test_size)
+            or ((train_end or val_start) and (val_end or test_start))
+        ):
+            raise ValueError(
+                "Invalid split input."
+                "You should specify either (`val_size` and `test_size`)"
+                " or ((`train_end` or `val_start`) and (`val_end` or `test_start`))"
+            )
+
+    def get_training_dataset(
```
jimdowling, line 890: Can you document the behaviour if there is no label defined on the FeatureView?

python/hsfs/feature_view.py
```diff
@@ -418,46 +415,583 @@ def create_training_dataset(
...
+        """
+        td, df = self._feature_view_engine.get_training_data(
+            self,
+            read_options,
+            training_dataset_version=training_dataset_version,
+            splits=[TrainingDatasetSplit.TRAIN, TrainingDatasetSplit.TEST],
+        )
+        return df[TrainingDatasetSplit.TRAIN] + df[TrainingDatasetSplit.TEST]
+
+    def get_train_validation_test_splits(
```
jimdowling, line 954: I don't have a problem with splits here (instead of 'split'). 
It is more correct english.

jimdowling (changes requested):

jimdowling:
I tried out the train_test_split(0.2) method, and validated against a dataset in a spreadsheet, and it matches the labels and features correctly, based on a few point-tests I made.

jimdowling:
Also, for 
x,y,x_test,y_test = fv.train_test_splits(0.0)
gives an error.
I would prefer if it returned empty x_test and y_test dataframes. 
But, you should document the expected behavior.
If test_size==0.0, you get an error, for example.

python/hsfs/core/feature_view_engine.py
```diff
@@ -163,42 +149,41 @@ def create_training_dataset(
...
+                feature_view_obj, training_dataset_version
+            )
+        else:
             td_updated = self._create_training_data_metadata(
                 feature_view_obj, training_dataset_obj
             )
+        # check splits
+        if len(splits) != len(td_updated.splits):
+            if len(td_updated.splits) == 0:
+                method_name = "get_training_dataset"
```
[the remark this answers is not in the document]
kennethmhc: I thought we use `training_data` for metadata + training dataset, while training dataset is  train_set, test_set, and val_set.

python/hsfs/feature_view.py
kennethmhc: How about set `y` to `None`?

PR #667, opened 2022-06-12
set hsfs client (hopsworks or external) automatically

davitbzh:


moritzmeister (changes requested):
Can you just change the docstrings and describe that by default it will set the flag to True if you are running HSFS outside of hopsworks? and else false?

java/src/main/java/com/logicalclocks/hsfs/engine/VectorServer.java
```diff
@@ -37,6 +38,8 @@
 import java.util.TreeMap;
 import java.util.stream.Collectors;
 
+import static com.logicalclocks.hsfs.metadata.HopsworksClient.getInstance;
```
moritzmeister, line 41: You have HopsworksClient already imported. You can just call `HopsworksClient.getInstance()` instead of importing the function again separately.

moritzmeister (changes requested):
nitpick

SirOibaf (approved):

PR #678, opened 2022-06-16
Set default external environment for python kernel

kennethmhc:


python/hsfs/engine/python.py
```diff
@@ -89,7 +89,12 @@ def _sql_offline(self, sql_query, feature_store, dataframe_type):
     def _jdbc(self, sql_query, connector, dataframe_type, read_options):
         if self._mysql_online_fs_engine is None:
             self._mysql_online_fs_engine = util.create_mysql_engine(
-                connector, "external" in read_options and read_options["external"]
+                connector,
+                (
+                    "external" in read_options
+                    and read_options["external"]
+                    or isinstance(client.get_instance(), client.external.Client)
```
SirOibaf, line 96: Doing it like this it would not allow you to overwrite the "external" flag if you are running out side Hopsworks.
We want to be able to overwrite it even if you are running outside Hopsworks for production use cases.

SirOibaf (changes requested):

python/hsfs/engine/python.py
kennethmhc: @SirOibaf alright! fixed.

PR #693, opened 2022-06-23
HOPSWORKS-3225 change throwing error to warning  while statistics on empty dataframe and add JDBC_FORMAT

DhananjayMukhedkar:
- change throwing warning instead of error while running statistics on empty dataframe
- add missing variable `JDBC_FORMAT` in RedshiftConnector

### DIFF
--- a/python/hsfs/core/statistics_engine.py
+++ b/python/hsfs/core/statistics_engine.py
@@ -58,12 +58,13 @@ class StatisticsEngine:
             commit_time = int(float(datetime.datetime.now().timestamp()) * 1000)
 
             content_str = self.profile_statistics(metadata_instance, feature_dataframe)
-            stats = statistics.Statistics(
-                commit_time=commit_time,
-                content=content_str,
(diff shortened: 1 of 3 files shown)


python/hsfs/core/statistics_engine.py
```diff
@@ -70,10 +71,11 @@ def compute_statistics(
     @staticmethod
     def profile_statistics(metadata_instance, feature_dataframe):
         if len(feature_dataframe.head(1)) == 0:
-            raise exceptions.FeatureStoreException(
+            warnings.warn(
                 "There is no data in the entity that you are trying to compute "
                 "statistics for. A possible cause might be that you inserted only data "
-                "to the online storage of a feature group."
+                "to the online storage of a feature group.",
+                category=util.StatisticsWarning,
```
SirOibaf, line 78: We need to return None here, otherwise the statistics computation will go ahead.

SirOibaf (changes requested):

SirOibaf (approved):

PR #697, opened 2022-07-01
Enable hudi multi write concurrency control

davitbzh:


SirOibaf:
Closing this as it's not relevant anymore. Hudi now supports file based concurrency control which is more suited for our use case: https://hudi.apache.org/docs/concurrency_control

PR #700, opened 2022-07-02
[HOPSWORKS-3235] Use Hudi snapshot query when users doesn't specify a…

SirOibaf:
…s_of in the query param


moritzmeister (approved):
Looks reasonable to me.

PR #705, opened 2022-07-07
[HOPSWORKS 3241] add progress bar during stream data ingestion in python

DhananjayMukhedkar:


python/hsfs/engine/python.py
```diff
@@ -751,9 +752,23 @@ def _write_dataframe_kafka(
...
-            if err is not None:
+            if err is not None and offline_write_options.get("debug_kafka", False):
                 print("Failed to deliver message: %s: %s" % (str(msg), str(err)))
-
+            else:
+                try:
+                    if progress_bar is not None:
+                        progress_bar.update()
+                except Exception as e:
+                    print("Failed tp update progress bar: {}".format(e))
```
robzor92, line 762:
```suggestion
                    print("Failed to update progress bar: {}".format(e))
```

robzor92 (changes requested):

python/hsfs/engine/python.py
```diff
@@ -751,9 +752,23 @@ def _write_dataframe_kafka(
         writer = self._get_encoder_func(feature_group._get_encoded_avro_schema())
 
         def acked(err, msg):
-            if err is not None:
+            if err is not None and offline_write_options.get("debug_kafka", False):
                 print("Failed to deliver message: %s: %s" % (str(msg), str(err)))
-
+            else:
+                try:
+                    if progress_bar is not None:
```
robzor92, line 759: I think you can skip this try catch block, or are you expecting a specific error?

robzor92 (changes requested):

moritzmeister (commented):
@robzor92 do we have to pin the tqdm version? probably not, right, so we are more flexible and it should't be as critical as other libraries.

robzor92:
> 

Not that I'm aware of, but I don't think so, probably some really old version will not be compatible but then alot of other libraries break too probably

python/hsfs/engine/python.py
```diff
@@ -751,9 +752,20 @@ def _write_dataframe_kafka(
         writer = self._get_encoder_func(feature_group._get_encoded_avro_schema())
 
         def acked(err, msg):
-            if err is not None:
+            if err is not None and offline_write_options.get("debug_kafka", False):
                 print("Failed to deliver message: %s: %s" % (str(msg), str(err)))
-
+            else:
+                if progress_bar is not None:
```
robzor92, line 758: I don't think we need these `is not None` checks, if it's none then it failed during initialization of the progress bar anyway.

python/hsfs/engine/python.py
```diff
@@ -806,6 +816,8 @@ def acked(err, msg):
 
         # make sure producer blocks and everything is delivered
         producer.flush()
+        if progress_bar is not None:
```
robzor92, line 819: I don't think we need these `is not None` checks, if it's none then it failed during initialization of the progress bar anyway.

robzor92 (changes requested):

python/hsfs/engine/python.py
DhananjayMukhedkar: agree, done

python/hsfs/engine/python.py
DhananjayMukhedkar: agree, done

robzor92 (approved):

PR #707, opened 2022-07-12
[HOPSWORKS-3202] Add filter to td

kennethmhc:
changes:
1. Fix query parameter in feature view API
2. Add extra filter to all get/create TD
3. Add td version to get batch query/data.
4. Add extra filter to td entity.
5. Remove default td version = 1 when init batch scoring since backend will not be able to fetch if there is none.

### DIFF
--- a/java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
@@ -65,6 +65,7 @@ public class FeatureView {
 
   private static FeatureViewEngine featureViewEngine = new FeatureViewEngine();
   private static VectorServer vectorServer = new VectorServer();
+  private Integer extraFilterVersion = null;
 
   public static class FeatureViewBuilder {
 
@@ -138,6 +139,10 @@ public class FeatureView {
     vectorServer.initServing(this, batch, external);
   }
 
+  public void initBatchScoring(Integer trainingDatasetVersion) {
+    this.extraFilterVersion = trainingDatasetVersion;
+  }
+
   @JsonIgnore
   public List<Object> getFeatureVector(Map<String, Object> entry)
       throws SQLException, FeatureStoreException, IOException, ClassNotFoundException {
@@ -170,41 +175,29 @@ public class FeatureView {
   @JsonIgnore
   public String getBatchQuery(String startTime, String endTime)
       throws FeatureStoreException, IOException, ParseException {
-    return getBatchQuery(startTime, endTime, null);
-  }
-
-  @JsonIgnore
-  public String getBatchQuery(String startTime, String endTime, Integer trainingDataVersion)
-      throws FeatureStoreException, IOException, ParseException {
     return featureViewEngine.getBatchQueryString(
         this,
         startTime != null ? FeatureGroupUtils.getDateFromDateString(startTime) : null,
         endTime != null ? FeatureGroupUtils.getDateFromDateString(endTime) : null,
-        trainingDataVersion);
+        extraFilterVersion);
   }
 
   @JsonIgnore
   public Dataset<Row> getBatchData(String startTime, String endTime)
       throws FeatureStoreException, IOException, ParseException {
-    return getBatchData(startTime, endTime, Maps.newHashMap(), null);
+    return getBatchData(startTime, endTime, Maps.newHashMap());
   }
 
   @JsonIgnore
-  public Dataset<Row> getBatchData(String startTime, String endTime, Integer trainingDataVersion)
-      throws FeatureStoreException, IOException, ParseException {
-    return getBatchData(startTime, endTime, Maps.newHashMap(), trainingDataVersion);
-  }
-
-  public Dataset<Row> getBatchData(String startTime, String endTime, Map<String, String> readOptions,
-      Integer trainingDataVersion)
+  public Dataset<Row> getBatchData(String startTime, String endTime, Map<String, String> readOptions)
       throws FeatureStoreException, IOException, ParseException {
     return featureViewEngine.getBatchData(
         this,
         startTime != null ? FeatureGroupUtils.getDateFromDateString(startTime) : null,
         endTime != null ? FeatureGroupUtils.getDateFromDateString(endTime) : null,
         readOptions,
-        trainingDataVersion);
-
+        extraFilterVersion
+    );
   }
 
   /**
--- a/java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
@@ -178,6 +178,12 @@ public class FeatureView {
   }
 
   @JsonIgnore
+  public Dataset<Row> getBatchData(String startTime, String endTime)
+      throws FeatureStoreException, IOException, ParseException {
+    return getBatchData(startTime, endTime, Maps.newHashMap(), null);
+  }
+
+  @JsonIgnore
   public Dataset<Row> getBatchData(String startTime, String endTime, Integer trainingDataVersion)
       throws FeatureStoreException, IOException, ParseException {
     return getBatchData(startTime, endTime, Maps.newHashMap(), trainingDataVersion);
--- a/java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
@@ -168,6 +168,16 @@ public class FeatureView {
   }
 
   @JsonIgnore
+  public String getBatchQuery(String startTime, String endTime)
+      throws FeatureStoreException, IOException, ParseException {
+    return featureViewEngine.getBatchQueryString(
+        this,
+        startTime != null ? FeatureGroupUtils.getDateFromDateString(startTime) : null,
+        endTime != null ? FeatureGroupUtils.getDateFromDateString(endTime) : null,
+        null);
+  }
+
+  @JsonIgnore
   public String getBatchQuery(String startTime, String endTime, Integer trainingDataVersion)
       throws FeatureStoreException, IOException, ParseException {
     return featureViewEngine.getBatchQueryString(
--- a/java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
@@ -616,12 +616,12 @@ public class FeatureView {
   }
 
   private void validateTrainTestSplit(Float testSize, String trainEnd, String testStart) throws FeatureStoreException {
-    if (!((testSize != null && testSize > 0)
+    if (!((testSize != null && testSize > 0 && testSize < 1)
         || (!Strings.isNullOrEmpty(trainEnd) || !Strings.isNullOrEmpty(testStart)))) {
       throw new FeatureStoreException(
           "Invalid split input."
               + "You should specify either `testSize` or (`trainEnd` or `testStart`)."
-              + " `testSize` should be greater than 0 if specified"
+              + " `testSize` should be between 0 and 1 if specified."
       );
     }
   }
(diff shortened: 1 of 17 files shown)


java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
```diff
@@ -183,31 +185,34 @@ public List<List<Object>> previewFeatureVectors(Integer n, boolean external)
...
   @JsonIgnore
-  public Dataset<Row> getBatchData(String startTime, String endTime)
+  public Dataset<Row> getBatchData(String startTime, String endTime, Integer trainingDataVersion)
       throws FeatureStoreException, IOException, ParseException {
-    return getBatchData(startTime, endTime, Maps.newHashMap());
+    return getBatchData(startTime, endTime, Maps.newHashMap(), trainingDataVersion);
   }
 
-  public Dataset<Row> getBatchData(String startTime, String endTime, Map<String, String> readOptions)
+  public Dataset<Row> getBatchData(String startTime, String endTime, Map<String, String> readOptions,
```
jimdowling, line 207: One thing about filters for training data is that they should only be applied when creating the training data. 
When i get batch data for scoring, often i don't want the filters to be applied.
If I don't supply the TD_Version here, will I get all the batch data in the time interval?

jimdowling: I have the surfing example, where i can walk through the use case with you to explain why i only want the filter on the TD data, not on the batch data for scoring.

python/hsfs/feature_view.py
```diff
@@ -161,7 +154,14 @@ def get_batch_query(
             `str`: batch query
         """
         return self._feature_view_engine.get_batch_query_string(
-            self, start_time, end_time
+            self,
+            start_time,
+            end_time,
+            training_dataset_version=(
```
jimdowling, line 160: Again, the TD_version should be an optional parameter for get_batch_data.

java/src/main/java/com/logicalclocks/hsfs/metadata/FeatureViewApi.java
```diff
@@ -292,15 +292,17 @@ public void deleteTrainingDatasetOnly(FeatureStore featureStore, String name, In
   }
 
   public Query getBatchQuery(FeatureStore featureStore, String name, Integer version,
-      Long startTime, Long endTime, Boolean withLabels) throws FeatureStoreException, IOException {
+      Long startTime, Long endTime, Boolean withLabels, Integer trainingDataVersion) throws FeatureStoreException,
```
jimdowling, line 295: Optional TD version.

jimdowling (changes requested):
I mentioned in the comments, for get_batch_data(), it should be optional to supply the TD_Version. In the only use case I have, i don't want to supply the TD_version on get_batch_data() - the surfing example. In the other theoretical examples we had before, such as "models for users from US/EU/Asia", we might want to batch score separately, and therefore apply the TD_version in get_batch_data(). I suspect we need both - where I would have it off by default and only if user supplies the TD_version, apply it.

java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
kennethmhc: > If I don't supply the TD_Version here, will I get all the batch data in the time interval?
yes, td version is optional in the current setting, you can pass `null` here.
> explain why i only want the filter on the TD data, not on the batch data for scoring.
To do this, we need to introduce a new flag "ignore_td_filter" because by default, when td contains statistics as well. If you do not provide td version, no transformation will occur. But I think having the extra may not be necessary. In your case, you can create a new td without filter, and get batch data from there.

jimdowling: Good point!
We would still want to apply the transformations and use the TD_version statistics. 
So a new flag - ignore_td_filter, would do the job. Is there a limit to 1 filter or can you add many filters? (ignore_td_filters)

We talked at the feature monitoring meeting last week about having a separate way to log the untransformed features - i know it's not directly related, but you might want to add an explicit flag for that - ignore_transformations.

jimdowling: Passing null as a parameter doesn't feel right though. Can we make it an optional parameter in Python (In java, it's not as big a deal, IMO).

java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
```diff
@@ -183,31 +185,34 @@ public List<List<Object>> previewFeatureVectors(Integer n, boolean external)
 
   @JsonIgnore
   public String getBatchQuery() throws FeatureStoreException, IOException, ParseException {
-    return getBatchQuery(null, null);
+    return getBatchQuery(null, null, null);
   }
 
   @JsonIgnore
-  public String getBatchQuery(String startTime, String endTime)
+  public String getBatchQuery(String startTime, String endTime, Integer trainingDataVersion)
```
bubriks, line 192: maybe a new method could be added that does not require `trainingDataVersion` (in a sense proxy method). This could be applicable in other places as well.

java/src/main/java/com/logicalclocks/hsfs/engine/FeatureViewEngine.java
```diff
@@ -170,43 +167,44 @@ public TrainingDatasetBundle getTrainingDataset(
...
-            featureView.getLabels());
-        computeStatistics(featureView, trainingDatasetUpdated, datasets);
-      } else {
-        trainingDatasetBundle = new TrainingDatasetBundle(trainingDatasetUpdated.getVersion(), dataset,
-            featureView.getLabels());
-        computeStatistics(featureView, trainingDatasetUpdated, new Dataset[]{dataset});
-      }
-      return trainingDatasetBundle;
-    }
+//    if (!IN_MEMORY_TRAINING_DATASET.equals(trainingDatasetUpdated.getTrainingDatasetType())) {
```
bubriks, line 170: why did it have to be commented out? why not just remove it if it is not necessary?

bubriks (commented):
overall looks good, just minor comments/questions

python/hsfs/feature_view.py
```diff
@@ -161,7 +154,14 @@ def get_batch_query(
             `str`: batch query
         """
         return self._feature_view_engine.get_batch_query_string(
-            self, start_time, end_time
+            self,
```
tdoehmen, line 157: New parameter still needs to be added to the docstring,

tdoehmen (commented):
Merge conflicts with master still need to be solved

python/hsfs/core/statistics_engine.py
```diff
@@ -78,7 +78,7 @@ def profile_statistics(metadata_instance, feature_dataframe):
                 "to the online storage of a feature group.",
                 category=util.StatisticsWarning,
             )
-            return None
+            return "{}"
```
tdoehmen, line 81: Is this change related to the PR`?

python/hsfs/feature_view.py
kennethmhc: Do you mean td version? It is not required in the parameter.

python/hsfs/core/statistics_engine.py
kennethmhc: I had an error when testing this PR. If the dataframe is empty, it causes error in the backend.

java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
```diff
@@ -162,7 +164,7 @@ public List<List<Object>> getFeatureVectors(Map<String, List<Object>> entry, boo
 
   @JsonIgnore
   public String getBatchQuery() throws FeatureStoreException, IOException, ParseException {
-    return getBatchQuery(null, null);
+    return getBatchQuery(null, null, null);
```
bubriks, line 167: no need for this change

java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
```diff
@@ -171,22 +173,41 @@ public String getBatchQuery(String startTime, String endTime)
     return featureViewEngine.getBatchQueryString(
```
bubriks, line 173: instead of calling `featureViewEngine.getBatchQueryString` could be better to call `this.getBatchQuery` to avoid duplication.

java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
```diff
@@ -382,14 +410,17 @@ public Integer createTrainValidationTestSplit(
...
+            .extraFilter(extraFilter)
             .build();
     return featureViewEngine.createTrainingDataset(this, trainingDataset, writeOptions).getVersion();
   }
 
   private List<Dataset<Row>> getDataset(TrainingDatasetBundle trainingDatasetBundle, List<String> splits) {
-    return splits.stream()
-        .flatMap(split -> trainingDatasetBundle.getDataset(split, true).stream())
-        .collect(Collectors.toList());
+    return null;
```
bubriks, line 420: why was this replaced?

java/src/main/java/com/logicalclocks/hsfs/constructor/Filter.java
```diff
@@ -19,8 +19,9 @@
 import com.logicalclocks.hsfs.Feature;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
+import lombok.NoArgsConstructor;
 import lombok.Setter;
-
+@NoArgsConstructor
```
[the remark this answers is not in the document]
bubriks: this issue looks to persist

bubriks (commented):
minor comments

bubriks:
also python lint fails

bubriks (approved):
Looks good to me.

tdoehmen:
Checkstyle fails in java client

[ERROR] C:\Users\doehmen\Documents\Hopsworks\feature-store-api\java\src\main\java\com\logicalclocks\hsfs\engine\FeatureViewEngine.java:238:73: '&&' should be on a new line. [OperatorWrap]
[ERROR] C:\Users\doehmen\Documents\Hopsworks\feature-store-api\java\src\main\java\com\logicalclocks\hsfs\engine\FeatureViewEngine.java:239:51: '&&' should be on a new line. [OperatorWrap]
[ERROR] C:\Users\doehmen\Documents\Hopsworks\feature-store-api\java\src\main\java\com\logicalclocks\hsfs\engine\FeatureViewEngine.java:243:73: '&&' should be on a new line. [OperatorWrap]
[ERROR] C:\Users\doehmen\Documents\Hopsworks\feature-store-api\java\src\main\java\com\logicalclocks\hsfs\engine\FeatureViewEngine.java:244:50: '&&' should be on a new line. [OperatorWrap]
Audit done.

python/hsfs/feature_view.py
tdoehmen: I was anticipating that you are going to add the td version parameter to that function, based on Jim's comment and the fact that the Java API getBatchQuery method has that version parameter. I don't know what the conclusion was between you and Jim was, but I think the API should be consistent between Java and Python.

python/hsfs/core/statistics_engine.py
tdoehmen: Was this an interegration test that failed?

jimdowling (approved):

python/hsfs/training_dataset.py
```diff
@@ -157,6 +161,8 @@ def __init__(
                 else extra_filter
             )
         else:
+            self._start_time = event_start_time
```
kennethmhc, line 160: `event_start_time` and `event_end_time` do not need to transformed if they are returned from backend.

python/hsfs/engine/spark.py
```diff
@@ -854,14 +854,14 @@ def save_empty_dataframe(self, feature_group, dataframe):
             {},
         )
 
-    def _apply_transformation_function(self, training_dataset, dataset):
+    def _apply_transformation_function(self, transformation_functions, dataset):
```
kennethmhc, line 857: only transformation functions is sufficient here.

python/hsfs/feature_view.py
```diff
@@ -132,7 +132,7 @@ def init_batch_scoring(
         """Initialise and cache parametrized transformation functions.
 
         # Arguments
-            training_dataset_version: int, optional. Default to be 1. Transformation statistics
+            training_dataset_version: int, optional. Default to be None. Transformation statistics
```
kennethmhc, line 135: TD version cannot be default to 1 here because the backend will try to retrieve the td metadata and get the extra filter. If we set the default to 1, it will cause 404 error. @davitbzh any comments?

tdoehmen:
When passing the extra filter in the Java API, the server complains:

`java.io.IOException: Error: 422{"type":"restApiJsonResponse","errorCode":120001,"errorMsg":"An argument was not provided or it was malformed.","usrMsg":"2022-01-06"}`

````
val filter = fv.getQuery().getLeftFeatures().get(0).gt("2022-01-06")
val ret = fv.createTrainingData(null, null, "transactions fraud batch training dataset", DataFormat.CSV,  true, null,null,null,null,null,null,filter);

tdoehmen:
All looks good. As soon as the tests passt I think we can merge

kennethmhc:
@tdoehmen I have fixed the tests.

PR #718, opened 2022-07-15
[HOPSWORKS-3250] Add serving key

kennethmhc:


python/hsfs/training_dataset.py
```diff
@@ -917,3 +917,15 @@ def test_end(self):
     @test_end.setter
     def test_end(self, test_end):
         self._test_end = test_end
+
+    @property
+    def serving_keys(self):
```
kennethmhc, line 922: Exist in the old API.

python/hsfs/feature_view.py
```diff
@@ -1227,3 +1227,16 @@ def schema(self):
...
         self._features = features
+
+    @property
+    def primary_keys(self):
+        """Set of primary key names that is used as keys in input dict object for `get_serving_vector` method."""
+        _vector_server = self._single_vector_server or self._batch_vectors_server
+        if _vector_server:
+            return _vector_server.serving_keys
+        else:
+            _vector_server = vector_server.VectorServer(
```
kennethmhc, line 1238: Use a new `VectorServer` to avoid `init_serving` because only `init_prepared_statement` is sufficient for returning primary keys.

moritzmeister (approved):
Looks good, the one on training datasets shouldn't be used anymore in the future I suppose, but good that you added it back.

PR #727, opened 2022-07-25
[HOPSWORKS-3260] Add python client tests

bubriks:


bubriks:
this includes changes to how imports are conducted, to accommodate testing

tdoehmen:
Hi, great effort on testing :) Just a question. What is reasoning behind adding the `python` prefix to the hsfs package name?

bubriks:
@tdoehmen The reason for addition of python is that otherwise tests look for files in the wrong paths. So in a sense addition of `python.` makes a local path into global.

moritzmeister:
> @tdoehmen The reason for addition of python is that otherwise tests look for files in the wrong paths. So in a sense addition of `python.` makes a local path into global.

I have the strong feeling with this you are just treating a symptom, but it's just something that you have to configure when running the tests. I haven't seen this pattern before, and it seems unnecessarily complicated.

Unfortunately, I wasn't here last week when we started this work, but did you guys discuss about using PyTest over unittest? I think using pytest will greatly reduce the boilerplate to write tests in the long run. From my experience, handling fixtures is much easier in pytest which we will need when we write tests using pyspark in local mode.

PR #741, opened 2022-07-27
[HOPSWORKS-3260] Add python client tests

bubriks:


tdoehmen (approved):
Thanks @bubriks for the immense effort. Looks good to me.

PR #745, opened 2022-07-28
Add unit testing action

moritzmeister:


bubriks (approved):
looks good, only issue on my part is that the tests covers code outside the specific method and not checking if API class is called (while not crucial here it could be important to know in other situations where conditions could apply)

moritzmeister:
@bubriks added the call to check if the API has been called.

PR #747, opened 2022-08-03
Fix atttribute access in validation report repr

moritzmeister:
- Add ge validate util job
- fix attribute in repr


kennethmhc (approved):
LGTM!

PR #756, opened 2022-08-11
[HOPSWORKS-3296] Make session token duration configurable in HSFS for storage connectors

moritzmeister:


tdoehmen (approved):

moritzmeister:
We are making it configurable instead.

PR #759, opened 2022-08-16
Set mergeSchema option to true to handle appended columns

SirOibaf:
When we append a feature to a feature group, the new data being written
will have a different schema than the existing Parquet files.

The new files will have the additional column while the old Parquet
files will not.

This change make sure that the schemas of the different Parquet files is
merged to obtain the full schema with all the necessary features.

Keep in mind that in Hopsworks Feature Store we do not allow users to
remove columns from an existing feature group.

Also release 2.5.16

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>com.logicalclocks</groupId>
     <artifactId>hsfs</artifactId>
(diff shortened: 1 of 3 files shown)


tdoehmen (approved):
looks good to me!

PR #760, opened 2022-08-17
[HOPSWORKS-3278] add type to feature view dto

ErmiasG:


ErmiasG:
https://github.com/logicalclocks/hops-testing-ee/pull/544

moritzmeister (approved):

python/hsfs/client/exceptions.py
```diff
@@ -19,14 +19,18 @@ class RestAPIError(Exception):
     """REST Exception encapsulating the response object and url."""
 
     def __init__(self, url, response):
-        error_object = response.json()
+        try:
+            error_object = response.json()
```
moritzmeister, line 23: why was this needed?

ErmiasG: B/c it was failing if there is no body and not showing the error. Hsml does the same.

moritzmeister (approved):

PR #783, opened 2022-09-15
[HOPSWORKS-3326] Support datetime in hsfs python API

davitbzh:
This PR adds option to provide python datetime object to date filters, such as: 
1. start and end times to features views and training datasets 
2. query such as `as_of` method

all timestamp conversion mothods were moved to `util.convert_event_time_to_timestamp`

JIRA Issue: https://hopsworks.atlassian.net/browse/HOPSWORKS-3326

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [x] Unit Tests
- [ ] Integration Tests
- [x] Manual Tests on VM

```
```


python/hsfs/util.py
```diff
@@ -139,6 +140,26 @@ def get_hudi_datestr_from_timestamp(timestamp):
     return datetime.fromtimestamp(timestamp / 1000).strftime("%Y%m%d%H%M%S%f")[:-3]
 
 
+def convert_event_time_to_timestamp(event_time, time_zone=timezone.utc):
+    if event_time is None:
+        return None
+    if isinstance(event_time, str):
+        if event_time:
+            return get_timestamp_from_date_string(event_time, time_zone)
+    if isinstance(event_time, pd._libs.tslibs.timestamps.Timestamp):
```
tdoehmen, line 149: Can we add support for `datetime.date` and `datetime.datetime` here as well?
For `datetime.date` we can assume the timezone given in the `time_zone `parameter given to `convert_event_time_to_timestamp.`
For date.datetime we can do the same when no tzinfo is set. If the datetime has tzinfo set, we convert it sensitive of that timezone.

python/hsfs/feature_view.py
```diff
@@ -219,10 +219,10 @@ def get_feature_vectors(
 
     def get_batch_data(self, start_time=None, end_time=None, read_options=None):
```
tdoehmen, line 220: There is a also a `get_batch_query` function which takes date arguments that has not been updated

`def get_batch_query(
        self, start_time: Optional[datetime] = None, end_time: Optional[datetime] = None
    )`

python/hsfs/feature_view.py
```diff
@@ -219,10 +219,10 @@ def get_feature_vectors(
...
         """
-        start_time: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-            following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`, or `%Y%m%d%H%M%S%f`.
-        end_time: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-            following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`,  or `%Y%m%d%H%M%S%f`.
+        start_time: timestamp in second or wallclock_time: Datetime object or string. The String should be formatted in
```
tdoehmen, line 222: I don't see a call of the timestamp conversion in the call trace of 

fv.get_batch_data()->fv-engine.get_batch_data()->fv-engine.get_batch_query()->fv-api->get_batch_query()

python/hsfs/feature_view.py
```diff
@@ -281,10 +281,12 @@ def create_training_data(
 
 
         # Arguments
-            start_time: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-                following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`, or `%Y%m%d%H%M%S%f`.
-            end_time: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-                following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`,  or `%Y%m%d%H%M%S%f`.
+            start_time: timestamp in second or wallclock_time: Datetime object or string. The String should be
```
tdoehmen, line 284: type hints for `start_time` and `end_time` were not updated

python/hsfs/feature_view.py
```diff
@@ -281,10 +281,12 @@ def create_training_data(
...
+                formatted in one of the following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`,
+                or `%Y%m%d%H%M%S%f`.
+            end_time: timestamp in second or wallclock_time: Datetime object or string. The String should be
+                formatted in one of the following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`,
+                or `%Y%m%d%H%M%S%f`.
```
tdoehmen, line 289: again, `untils.convert_event_time_to_timestamp` does not seem to be called on the input

python/hsfs/feature_view.py
```diff
@@ -387,14 +389,18 @@ def create_train_test_split(
...
-            train_start: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-                following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`, or `%Y%m%d%H%M%S%f`.
-            train_end: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-                following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`,  or `%Y%m%d%H%M%S%f`.
-            test_start: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-                following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`, or `%Y%m%d%H%M%S%f`.
-            test_end: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-                following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`,  or `%Y%m%d%H%M%S%f`.
+            train_start: timestamp in second or wallclock_time: Datetime object or string. The String should be
+                formatted in one of the following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`,
```
tdoehmen, line 393: again, type hints for function parameters were not updated to include timestamp, datetime etc.

python/hsfs/feature_view.py
```diff
@@ -509,18 +515,24 @@ def create_train_validation_test_split(
...
-            validation_start: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-                following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`, or `%Y%m%d%H%M%S%f`.
-            validation_end: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-                following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`,  or `%Y%m%d%H%M%S%f`.
-            test_start: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-                following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`, or `%Y%m%d%H%M%S%f`.
-            test_end: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-                following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`,  or `%Y%m%d%H%M%S%f`.
+            train_start: timestamp in second or wallclock_time: Datetime object or string. The String should be
+                formatted in one of the following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`,
```
tdoehmen, line 519: Update type hints..

python/hsfs/feature_view.py
```diff
@@ -720,14 +734,18 @@ def train_test_split(
...
-            train_start: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-                following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`, or `%Y%m%d%H%M%S%f`.
-            train_end: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-                following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`,  or `%Y%m%d%H%M%S%f`.
-            test_start: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-                following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`, or `%Y%m%d%H%M%S%f`.
-            test_end: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-                following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`,  or `%Y%m%d%H%M%S%f`.
+            train_start: timestamp in second or wallclock_time: Datetime object or string. The String should be
+                formatted in one of the following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`,
```
tdoehmen, line 738: update type hints...

python/hsfs/feature_view.py
```diff
@@ -818,18 +836,24 @@ def train_validation_test_split(
...
+                or `%Y%m%d%H%M%S%f`.
+            train_end: timestamp in second or wallclock_time: Datetime object or string. The String should be
+                formatted in one of the following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`,
+                or `%Y%m%d%H%M%S%f`.
```
tdoehmen, line 844: type hints...

python/hsfs/feature_view.py
```diff
@@ -650,10 +662,12 @@ def training_data(
         recreate the training data.
 
         # Arguments
-            start_time: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-                following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`, or `%Y%m%d%H%M%S%f`.
-            end_time: timestamp in second or wallclock_time: Datetime string. The String should be formatted in one of the
-                following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`,  or `%Y%m%d%H%M%S%f`.
+            start_time: timestamp in second or wallclock_time: Datetime object or string. The String should be
+                formatted in one of the following formats `%Y%m%d`, `%Y%m%d%H`, `%Y%m%d%H%M`, `%Y%m%d%H%M%S`,
+                or `%Y%m%d%H%M%S%f`.
```
tdoehmen, line 667: conversion is not called (?)

python/hsfs/util.py
```diff
@@ -139,6 +140,26 @@ def get_hudi_datestr_from_timestamp(timestamp):
     return datetime.fromtimestamp(timestamp / 1000).strftime("%Y%m%d%H%M%S%f")[:-3]
```
tdoehmen, line 142: I don't think we still actually need the `get_timestamp_from_date_string`. It would be nice if we could unify it with `convert_event_time_to_timestamp`.

python/hsfs/util.py
```diff
@@ -139,6 +140,26 @@ def get_hudi_datestr_from_timestamp(timestamp):
...
 
 
+def convert_event_time_to_timestamp(event_time, time_zone=timezone.utc):
+    if event_time is None:
+        return None
+    if isinstance(event_time, str):
+        if event_time:
```
tdoehmen, line 147: Why do we need this check here? Should be covered by the null check above.

python/hsfs/constructor/query.py
```diff
@@ -221,21 +221,18 @@ def as_of(self, wallclock_time=None, exclude_until=None):
...
-            util.get_timestamp_from_date_string(wallclock_time)
-            if wallclock_time
-            else None
-        )
+        wallclock_timestamp = util.convert_event_time_to_timestamp(wallclock_time)
+
         exclude_until_timestamp = (
-            util.get_timestamp_from_date_string(exclude_until)
+            util.convert_event_time_to_timestamp(exclude_until)
             if exclude_until
```
tdoehmen, line 236: We can exclude the null check here, because this is done in `util.convert_event_time_to_timestamp` now as well

python/hsfs/util.py
```diff
@@ -139,6 +140,26 @@ def get_hudi_datestr_from_timestamp(timestamp):
...
+            return get_timestamp_from_date_string(event_time, time_zone)
+    if isinstance(event_time, pd._libs.tslibs.timestamps.Timestamp):
+        # pandas.timestamp represents millisecond in decimal
```
tdoehmen, line 150: This comments sounds a bit misleading. pd.Timestamp.timestamp() returns unix epoch time in seconds and we multiply by 1000 because we want to return unix epoch time in milliseconds.

python/hsfs/core/feature_view_engine.py
```diff
@@ -119,8 +119,8 @@ def get_batch_query_string(
         query_obj = self._feature_view_api.get_batch_query(
             feature_view_obj.name,
             feature_view_obj.version,
-            start_time,
-            end_time,
+            util.convert_event_time_to_timestamp(start_time),
```
tdoehmen, line 122: Why is the inpur converted in `get_batch_query_string`, but not in `get_batch_query`?

python/hsfs/training_dataset.py
```diff
@@ -125,8 +124,8 @@ def __init__(
             self._training_dataset_type = None
         # set up depending on user initialized or coming from backend response
         if created is None:
-            self._start_time = self._convert_event_time_to_timestamp(event_start_time)
-            self._end_time = self._convert_event_time_to_timestamp(event_end_time)
+            self._start_time = util.convert_event_time_to_timestamp(event_start_time)
```
tdoehmen, line 127: Maybe we should do the conversion as well in `_set_time_splits`

python/hsfs/training_dataset.py
```diff
@@ -230,25 +229,11 @@ def _append_time_split(
...
-                return util.get_timestamp_from_date_string(event_time, timezone.utc)
-        elif isinstance(event_time, int):
-            if event_time == 0:
-                raise ValueError("Event time should be greater than 0.")
-            # jdbc supports timestamp precision up to second only.
-            return event_time * 1000
-        else:
-            raise ValueError("Given event time should be in `str` or `int` type")
-
     def save(
```
tdoehmen, line 237: `get_statistics` could have the conversion as well

python/tests/engine/test_python.py
```diff
@@ -1741,22 +1735,19 @@ def test_convert_to_unix_timestamp_str(self, mocker):
...
         # Assert
         assert result == 1483225200000
 
     def test_convert_to_unix_timestamp_int(self):
-        # Arrange
-        python_engine = python.Engine()
-
         # Act
-        result = python_engine._convert_to_unix_timestamp(t=1483225200)
+        result = util.convert_event_time_to_timestamp(event_time=1483225200)
```
tdoehmen, line 1750: I am pro a new test class `test_utils.py` where the new `util.convert_event_time_to_timestamp` gets tested with various inputs. The tests should cover all input types (int, string, timestamp, datetime.datetime, datetime.date), and in particular different supported and not supported string-based input formats, timestamps and datetimes with timezones, and other possible edge cases.

python/hsfs/feature_view.py
davitbzh: that method call `get_batch_query_string` and that method uses `convert_event_time_to_timestamp`

python/tests/engine/test_python.py
davitbzh: added couple of more tests in `test_python.py`

python/hsfs/feature_view.py
```diff
@@ -444,10 +452,10 @@ def create_train_test_split(
             version=None,
             test_size=test_size,
             time_split_size=2,
-            train_start=train_start,
-            train_end=train_end,
-            test_start=test_start,
-            test_end=test_end,
+            train_start=util.convert_event_time_to_timestamp(train_start),
```
kennethmhc, line 455: Raw `train_start` without conversion should be good enough because the conversion occurs in the constructor of TD.

python/hsfs/feature_view.py
```diff
@@ -576,12 +590,12 @@ def create_train_validation_test_split(
             validation_size=validation_size,
             test_size=test_size,
             time_split_size=3,
-            train_start=train_start,
-            train_end=train_end,
-            validation_start=validation_start,
-            validation_end=validation_end,
-            test_start=test_start,
-            test_end=test_end,
+            train_start=util.convert_event_time_to_timestamp(train_start),
```
kennethmhc, line 593: same here.

python/hsfs/feature_view.py
```diff
@@ -679,8 +695,8 @@ def training_data(
             name=self.name,
             version=None,
             splits={},
-            event_start_time=start_time,
-            event_end_time=end_time,
+            event_start_time=util.convert_event_time_to_timestamp(start_time),
```
kennethmhc, line 698: same here and in other td creation methods.

kennethmhc (changes requested):
2 tests in test_spark.py failed

python/hsfs/feature_view.py
kennethmhc: @davitbzh `util.convert_event_time_to_timestamp` is not necessary here.

python/hsfs/util.py
```diff
@@ -143,12 +142,21 @@ def get_hudi_datestr_from_timestamp(timestamp):
 def convert_event_time_to_timestamp(event_time, time_zone=timezone.utc):
```
tdoehmen, line 142: Since the time_zone argument is not used anywhere, I would argue for removing it here and to rely on the timzone information provided with the timestamp/datetime objects. Is there is no timezone information, e.g. when the user provides a string input, we can always set it to UTC.

python/hsfs/util.py
```diff
@@ -143,12 +142,21 @@ def get_hudi_datestr_from_timestamp(timestamp):
...
         if event_time:
             return get_timestamp_from_date_string(event_time, time_zone)
-    if isinstance(event_time, pd._libs.tslibs.timestamps.Timestamp):
-        # pandas.timestamp represents millisecond in decimal
+    elif isinstance(event_time, pd._libs.tslibs.timestamps.Timestamp):
+        # convert to unix epoch time in milliseconds.
         return event_time.timestamp() * 1000
+    elif isinstance(event_time, datetime):
+        # convert to unix epoch time in milliseconds.
+        return int(event_time.replace(tzinfo=time_zone).timestamp() * 1000)
```
tdoehmen, line 153: In the timezone PR here https://github.com/logicalclocks/feature-store-api/pull/789, I only set the timezone to UTC when `datetime.tzinfo` is None, otherwise I keep the timezone information. Let's do the same here.

python/hsfs/util.py
```diff
@@ -143,12 +142,21 @@ def get_hudi_datestr_from_timestamp(timestamp):
...
     if event_time is None:
         return None
-    if isinstance(event_time, str):
+    elif isinstance(event_time, str):
         if event_time:
             return get_timestamp_from_date_string(event_time, time_zone)
-    if isinstance(event_time, pd._libs.tslibs.timestamps.Timestamp):
-        # pandas.timestamp represents millisecond in decimal
+    elif isinstance(event_time, pd._libs.tslibs.timestamps.Timestamp):
+        # convert to unix epoch time in milliseconds.
```
tdoehmen, line 149: `pd.Timestamps` can also have timezone information. To keep the logic consistent between pd.Timestamp and `datetime.datetime` you can first convert the pd.Timestamp to a datetime object by calling `event_time.to_pydatetime()` (`https://pandas.pydata.org/docs/reference/api/pandas.Timestamp.to_pydatetime.html`) and then re-use the datetime logic.

python/tests/engine/test_python.py
```diff
@@ -1752,6 +1753,24 @@ def test_convert_to_unix_timestamp_int(self):
         # Assert
         assert result == 1483225200000
 
+    def test_convert_to_unix_timestamp_datetime(self):
+        # Act
+        result = util.convert_event_time_to_timestamp(
+            event_time=datetime(2022, 9, 18), time_zone=timezone.utc
```
tdoehmen, line 1759: A test for pd.Timestamp is still missing. In best case the tests would also cover a datetime/timestamps with tzinfo set. e.g. `datetime(2022, 7, 3, 0, 0, 0).replace(tzinfo=timezone(timedelta(hours=3)))` and `pd.Timestamp("2022-07-03T00", tz="US/Pacific")`

python/hsfs/util.py
tdoehmen: I'd also be fine with dropping pd.Timezone support. I see that in the typehints it's also not mentioned and maybe it doesn't add much value, since users can call .to_pydatetime() on pd.Timestamp object if required.

python/hsfs/util.py
tdoehmen: This could still be adressed if I'm not mistaken

python/hsfs/core/statistics_engine.py
```diff
@@ -176,7 +176,7 @@ def get(
         training_dataset_version=None,
     ):
         """Get Statistics with the specified commit time of an entity."""
```
tdoehmen, line 178: FeatureGroup also has a function `def get_statistics(self, commit_time: str = None):` which should be updated in a similar way.

python/hsfs/feature_group.py
```diff
@@ -1283,9 +1283,9 @@ def as_of(self, wallclock_time, exclude_until=None):
             when calling the `insert()` method.
 
         # Arguments
-            wallclock_time: Datetime string. The String should be formatted in one of the
+            wallclock_time: timestamp, Datetime object or string. The String should be formatted in one of the
```
tdoehmen, line 1286: One last comment about the docstrings. We should include all suppoerted types here, which are now:
`datatime.datetime, datetime.date, unix time in seconds (int), or string.` 
Also the docsstring shold be consistent across all parameters that have the `Optional[Union[str, int, datetime, date]]` input type. Right now, e.g. `FeatureView.train_validation_test_split` still has a different/old docstring.

tdoehmen: We also support timestamps is formats with spacing (e.g. `%Y-%m-%d`, `%Y-%m-%d %H`, `%Y-%m-%d %H:%M`, or `%Y-%m-%d %H:%M:%S`). But I don't think we can cover all the variants in the docstring of the function. Maybe we can document the possible variants in a guide and link there from the docstring (?)

tdoehmen: I created a Jira for this here https://hopsworks.atlassian.net/browse/FSTORE-316

python/hsfs/core/feature_view_engine.py
davitbzh: get_batch_query calls get_batch_query_string

python/hsfs/training_dataset.py
davitbzh: done

python/hsfs/training_dataset.py
davitbzh: done

python/hsfs/util.py
davitbzh: done

python/hsfs/util.py
davitbzh: done

python/hsfs/core/statistics_engine.py
davitbzh: this is converted in statistics engine `get` mothod

tdoehmen: Okay, but then the type hints need to be updated

davitbzh: done both for FG and TD

tdoehmen: Looks like this was done

tdoehmen:
Alright, looks good to me now!

kennethmhc (approved):
LGTM

PR #816, opened 2022-10-04
[FSTORE-346] Handle as_of in feature view

kennethmhc:
- Show warning when there is `as_of` in the query when creating a feature view
- Raise exception when reading incremental query from python engine

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [x] Unit Tests
- [ ] Integration Tests
- [x] Manual Tests on VM

```
```

### DIFF
--- a/java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
@@ -112,8 +112,8 @@ public class FeatureView {
     public FeatureViewBuilder query(Query query) {
       this.query = query;
       if (query.isTimeTravel()) {
-        LOGGER.info("`as_of` argument in the `Query` will be ignored because" +
-            " feature view does not support time travel query.");
+        LOGGER.info("`as_of` argument in the `Query` will be ignored because "
+            + "feature view does not support time travel query.");
       }
       return this;
     }
--- a/java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
@@ -16,6 +16,8 @@ import lombok.NonNull;
 import lombok.Setter;
 import org.apache.spark.sql.Dataset;
 import org.apache.spark.sql.Row;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.sql.SQLException;
@@ -70,6 +72,7 @@ public class FeatureView {
   private static FeatureViewEngine featureViewEngine = new FeatureViewEngine();
   private static VectorServer vectorServer = new VectorServer();
   private Integer extraFilterVersion = null;
+  private static final Logger LOGGER = LoggerFactory.getLogger(FeatureGroup.class);
 
   public static class FeatureViewBuilder {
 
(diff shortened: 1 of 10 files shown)


java/src/main/java/com/logicalclocks/hsfs/FeatureView.java
```diff
@@ -99,8 +102,19 @@ public FeatureViewBuilder description(String description) {
...
+     * not support time travel query.
+     *
+     * @param query
+     * @return builder
+     */
     public FeatureViewBuilder query(Query query) {
       this.query = query;
+      if (query.isTimeTravel()) {
+        LOGGER.info("`as_of` argument in the `Query` will be ignored because" +
+            " feature view does not support time travel query.");
```
tdoehmen, line 116: Stylecheck still wants to thave the `+` in the next line:
`C:\Users\doehmen\Documents\Hopsworks\feature-store-api\java\src\main\java\com\logicalclocks\hsfs\FeatureView.java:115:79: '+' should be on a new line. [OperatorWrap]`

python/hsfs/core/feature_view_engine.py
```diff
@@ -61,6 +62,11 @@ def __init__(self, feature_store_id):
         self._query_constructor_api = query_constructor_api.QueryConstructorApi()
 
     def save(self, feature_view_obj):
+        if feature_view_obj.query.is_time_travel():
```
tdoehmen, line 65: I think the check should be recursive for all sub-queries within the query. If, e.g., the right side of the join is a time-travel query we should catch that as well.

python/hsfs/constructor/query.py
```diff
@@ -237,6 +238,7 @@ def as_of(self, wallclock_time=None, exclude_until=None):
         # Returns
             `Query`. The query object with the applied time travel condition.
         """
+        self._is_time_travel = True
```
tdoehmen, line 241: setting `_is_time_travel` should be move at the end of the function. Otherwise it would set _it_time_travel to true, even if, e.g., `until.convert_event_time_to_timestamp` fails and time travel never gets activated.

python/hsfs/constructor/query.py
```diff
@@ -49,6 +49,7 @@ def __init__(
         self._storage_connector_api = storage_connector_api.StorageConnectorApi(
             feature_store_id
         )
+        self._is_time_travel = False
```
tdoehmen, line 52: For the sake of consistency, we should set _is_time_travel according to whether "left_feature_group_start_time" and "left_feature_group_start_time" are None/non None.

python/hsfs/constructor/query.py
tdoehmen: Also, if as_of gets called with both arguments `None`, we should not set `_is_time_travel=True`

python/hsfs/core/feature_view_engine.py
kennethmhc: Since we set `is_time_travel` from the top level, so it should be sufficient to check only the top level, right?

java/src/main/java/com/logicalclocks/hsfs/constructor/Query.java
```diff
@@ -303,4 +303,17 @@ public Query appendFeature(Feature feature) {
     this.leftFeatures.add(feature);
     return this;
   }
+
+  public boolean isTimeTravel() {
+    if (leftFeatureGroupStartTime != null || leftFeatureGroupEndTime != null) {
+      return true;
+    }
+    for (Join join: joins) {
+      if (join.getQuery().leftFeatureGroupStartTime != null
```
tdoehmen, line 312: Why not call `.isTimeTravel` on the join.getQuery(). I think that would also make it real recursive, as right now only the first level of joins in checked.

python/tests/engine/test_python.py
```diff
@@ -599,6 +600,40 @@ def test_register_hudi_temporary_table(self):
         # Assert
         assert result is None
 
+    def test_register_hudi_temporary_table_time_travel(self):
```
tdoehmen, line 603: And can you add one more small test which covers the recursive thing?

kennethmhc: I added 2 tests in test_feature_view_engine.

tdoehmen (approved):

PR #848, opened 2022-11-02
[FSTORE-328][APPEND] throw intelligible error message if user set primary and partition keys doesn't exist

davitbzh:
This PR fixes bug related to key attribute verification for ExternalFeatureGroups 

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [x] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


python/hsfs/util.py
```diff
@@ -233,33 +233,36 @@ def get_hostname_replaced_url(sub_path: str):
     return url_parsed.geturl()
 
 
-def verify_attribute_key_names(feature_group_obj):
+def verify_attribute_key_names(feature_group_obj, external_feature_group=False):
```
moritzmeister, line 236: This is unnecessary, you can just use `isinstance()` on the feature group object to check if it's an external feature group.

moritzmeister (changes requested):

moritzmeister (commented):
How come you don't need to distinguish between external and normal in the java part?

python/hsfs/util.py
davitbzh: I did that :) but importing FG in util class introduces circular import error

davitbzh:
> How come you don't need to distinguish between external and normal in the java part?

In Java we provide partition and precombine key names separately (I don´t remember what was the reason for it though). and then setting those values as null in externalFeatureGroupEngine.

moritzmeister (approved):

PR #852, opened 2022-11-08
[FSTORE-398] add validation history method to python client

vatj:
This PR adds the possibility to query the validation history associated with an expectation by calling fg.get_validation_history(expectationId)

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/keyFile.json
+++ b/keyFile.json
@@ -1,0 +1,1 @@
+{"type": "service_account", "project_id": "test", "private_key_id": "123456", "private_key": "-----BEGIN PRIVATE KEY-----test-----END PRIVATE KEY-----", "client_email": "test@project.iam.gserviceaccount.com"}
--- a/keyFile.json
+++ b/keyFile.json
@@ -1,1 +1,0 @@
-{"type": "service_account", "project_id": "test", "private_key_id": "123456", "private_key": "-----BEGIN PRIVATE KEY-----test-----END PRIVATE KEY-----", "client_email": "test@project.iam.gserviceaccount.com"}
--- a/python/hsfs/core/validation_result_engine.py
+++ b/python/hsfs/core/validation_result_engine.py
@@ -20,6 +20,7 @@ from datetime import datetime, date
 from hsfs.util import convert_event_time_to_timestamp
 
 from hsfs.ge_validation_result import ValidationResult
+from great_expectations.core import ExpectationValidationResult
 
 
 class ValidationResultEngine:
@@ -42,7 +43,8 @@ class ValidationResultEngine:
         end_validation_time: Union[str, int, datetime, date, None] = None,
         ingested_only: bool = False,
         rejected_only: bool = False,
-    ) -> Union[List[ValidationResult], ValidationResult]:
+        ge_type: bool = True,
+    ) -> Union[List[ValidationResult], List[ExpectationValidationResult]]:
         """Get Validation Results relevant to an Expectation specified by expectation_id.
 
         :param expectation_id: id of the expectation for which to fetch the validation history
@@ -51,9 +53,13 @@ class ValidationResultEngine:
         :type ingestion_only: bool
         :param rejected_only: retrieve only validation result linked to data not ingested in the Feature Group
         :type rejected_only: bool
-        :param start_validation_time: retrieve validation result posterior to start_validation_time
+        :param ge_type: whether to convert Hopsworks object to native Great Expectations object
+        :type ge_type: bool
+        :param start_validation_time: retrieve validation result posterior to start_validation_time.
+        Supported format include timestamps(int), datetime, date or string formatted to be datutils parsable.
         :type start_validation_time: Union[str, int, datetime, date, None]
         :param end_validation_time: retrieve validation result anterior to end_validation_time
+        Supported format include timestamps(int), datetime, date or string formatted to be datutils parsable.
         :type end_validation_time: Union[str, int, datetime, date, None]
         """
         query_params = self._build_query_params(
@@ -63,9 +69,17 @@ class ValidationResultEngine:
             end_validation_time=end_validation_time,
         )
 
-        return self._validation_result_api.get_validation_history(
+        history = self._validation_result_api.get_validation_history(
             expectation_id=expectation_id, query_params=query_params
         )
+
+        if isinstance(history, ValidationResult):
+            history = [history]
+
+        if ge_type:
+            return [result.to_ge_type() for result in history]
+        else:
+            return history
 
     def _build_query_params(
         self,
@@ -77,7 +91,7 @@ class ValidationResultEngine:
         query_params = {"filter_by": [], "sort_by": "validation_time:desc"}
 
         if ingested_only and rejected_only:
-            raise ValueError("ingested_only and rejected_only ")
+            raise ValueError("ingested_only and rejected_only cannot be both True.")
         elif ingested_only:
             query_params["filter_by"].append("ingestion_result_eq:ingested")
         elif rejected_only:
--- a/python/hsfs/core/validation_result_engine.py
+++ b/python/hsfs/core/validation_result_engine.py
@@ -1,0 +1,103 @@
+#
+#   Copyright 2022 Hopsworks AB
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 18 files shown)


python/hsfs/feature_group.py
```diff
@@ -535,6 +537,35 @@ def save_validation_report(
             return self._validation_report_engine.save(report).to_ge_type()
         return self._validation_report_engine.save(report)
 
+    def get_validation_history(
+        self,
+        expectation_id: int,
+        sort_by: Optional[str] = "validation_time:desc",
+        filter_by: Optional[str] = None,
+        offset: Optional[int] = None,
```
kennethmhc, line 545: Pagination is needed only in UI right?

python/hsfs/feature_group.py
```diff
@@ -535,6 +537,35 @@ def save_validation_report(
             return self._validation_report_engine.save(report).to_ge_type()
         return self._validation_report_engine.save(report)
 
+    def get_validation_history(
+        self,
+        expectation_id: int,
+        sort_by: Optional[str] = "validation_time:desc",
```
kennethmhc, line 543: Does it make sense if we always return a list sort by time in descending order? User can reverse the order easily in python.

python/hsfs/feature_group.py
```diff
@@ -535,6 +537,35 @@ def save_validation_report(
...
+        offset: Optional[int] = None,
+        limit: Optional[int] = None,
+    ) -> List[ValidationResult]:
+        """Fetch validation history of an Expectation specified by its id.
+
+        # Arguments
+            expectation_id: id of the Expectation for which to fetch the validation history
+            sort_by: sort the validation results according to validation time, descending or ascending. Value can be either
+                validation_time:desc or validation_time:asc
+            filter_by: filter the validation results based on validation time or ingestion result.
```
kennethmhc, line 554: How about hiding the implementation detail of the backend? We can have
```
ingested_only=False
rejected_only=False
# For the time we can reference to fg.read_changes
start_validation_time: Union[str, int, datetime, date],
end_validation_time: Union[str, int, datetime, date],
```
This seems more user friendly.

python/hsfs/core/validation_result_engine.py
```diff
@@ -0,0 +1,154 @@
...
+
+        return self._validation_result_api.get_validation_history(
+            expectation_id=expectation_id,
+            offset=offset,
+            limit=limit,
+            filter_by=filter_by,
+            sort_by=sort_by,
+        )
+
+    def _verify_sort_by(self, sort_by: str) -> None:
```
kennethmhc, line 77: After changing the API,  users do not pass sort_by and filter_by parameter by themselves. This also simplified the code because we do not need to verify them.

python/hsfs/ge_validation_result.py
```diff
@@ -174,10 +185,38 @@ def expectation_config(self, expectation_config):
                 "Expectation config field must be stringified json or dict"
             )
 
-    def __str__(self):
+    @property
+    def validation_time(self) -> Optional[datetime.datetime]:
+        return self._validation_time
+
+    @validation_time.setter
+    def validation_time(self, validation_time: Optional[int]) -> None:
```
kennethmhc, line 193: Can use `util.convert_event_time_to_timestamp` here to make handling of time in the API more consistent.

python/hsfs/ge_validation_result.py
```diff
@@ -174,10 +184,42 @@ def expectation_config(self, expectation_config):
...
 
-    def __str__(self):
+    @property
+    def validation_time(self) -> Optional[int]:
+        return self._validation_time
+
+    @validation_time.setter
+    def validation_time(
+        self, validation_time: Union[str, int, datetime, date, None]
+    ) -> None:
```
kennethmhc, line 194: A doc string describing different input type of validation_time would be helpful.

python/tests/core/test_validation_result_engine.py
```diff
@@ -0,0 +1,165 @@
...
+        # Second case
+        assert len(correct_outputs[1]["filter_by"]) == 2
+        assert "ingestion_result_eq:ingested" in correct_outputs[1]["filter_by"]
+        filter_validation_gte = [
+            val
+            for val in filter(
+                lambda x: "validation_time_gte:" in x, correct_outputs[1]["filter_by"]
+            )
+        ]
+        assert len(filter_validation_gte) == 1
```
kennethmhc, line 142: check `len(filter_validation_lte) == 0` as well?

kennethmhc (changes requested):
Looks good. Just few small comments.

python/hsfs/ge_validation_result.py
vatj: We don't want the user to set anything on these objects. They should really just be intermediate for us so I am not sure of the value but I will add something.

python/hsfs/core/validation_result_engine.py
```diff
@@ -0,0 +1,105 @@
...
+    ) -> Union[List[ValidationResult], ValidationResult]:
+        """Get Validation Results relevant to an Expectation specified by expectation_id.
+
+        :param expectation_id: id of the expectation for which to fetch the validation history
+        :type expectation_id: int
+        :param ingestion_only: retrieve only validation result linked to data ingested in the Feature Group
+        :type ingestion_only: bool
+        :param rejected_only: retrieve only validation result linked to data not ingested in the Feature Group
+        :type rejected_only: bool
+        :param start_validation_time: retrieve validation result posterior to start_validation_time
```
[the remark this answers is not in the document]
vatj: Ja I don't see any point in making them exclusive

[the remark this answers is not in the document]
kennethmhc: Would be nice to write it down as well.

python/hsfs/feature_group.py
```diff
@@ -678,11 +709,14 @@ def expectation_suite(
                 feature_group_id=self._id,
             )
         elif isinstance(expectation_suite, dict):
-            self._expectation_suite = ExpectationSuite(
-                **expectation_suite,
-                feature_store_id=self._feature_store_id,
-                feature_group_id=self._id,
-            )
+            if "feature_store_id" in expectation_suite.keys():
+                self._expectation_suite = ExpectationSuite(**expectation_suite)
```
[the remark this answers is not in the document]
kennethmhc: I mean if the expectation_suite are from a shared feature store, here should fs_id be the current fs or the shared fs?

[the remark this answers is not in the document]
vatj: Expectation suite is attached to the FG, it will be the same fs_id than the feature_group_id

python/hsfs/feature_group.py
```diff
@@ -1049,7 +1105,7 @@ def insert(
         operation: Optional[str] = "upsert",
         storage: Optional[str] = None,
         write_options: Optional[Dict[Any, Any]] = {},
-        validation_options: Optional[Dict[Any, Any]] = {},
+        validation_options: Optional[Dict[Any, Any]] = {"save_report": True},
```
moritzmeister, line 1108: This is potentially dangerous, cause the dict is defined when the function is defined the first time. So that means if the dict gets modified in a call to the function, it will stay modified for all future calls to the method.
This is a python thing, and we had problems with it in the past.

So please double check, if we modify it anywhere in the function call.

python/hsfs/util.py
```diff
@@ -102,48 +103,11 @@ def create_mysql_engine(online_conn, external):
     return sql_alchemy_engine
 
 
-def check_timestamp_format_from_date_string(input_date):
```
moritzmeister, line 105: What happened to this method? are you sure it's not used anywhere else?

python/hsfs/feature_group.py
vatj: I can change it to `Optional[Dict[Any, Any]] = None` and then:
```python3
if validation_options is None:
  validation_options = {'save_report': True}
```

moritzmeister: yes, let's do it, just to be sure

python/hsfs/util.py
vatj: dateutils parser covers all former date parsing + some and all the unit test for it passed.

moritzmeister: ok great

moritzmeister (approved):

keyFile.json
```diff
@@ -0,0 +1 @@
+{"type": "service_account", "project_id": "test", "private_key_id": "123456", "private_key": "-----BEGIN PRIVATE KEY-----test-----END PRIVATE KEY-----", "client_email": "test@project.iam.gserviceaccount.com"}
```
kennethmhc, line 1: Is this necessary?

vatj: No I think it is pytest that generates it everytime, don't understand why

kennethmhc (approved):
LGTM!

PR #854, opened 2022-11-08
[FSTORE-414] feature_view.create_train_test_split returns empty df

davitbzh:
This PR adds/fixes/changes...
Updated converting to epoch milliseconds in spark engine and use util function to be consistent 

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [x] Unit Tests
- [ ] Integration Tests
- [x] Manual Tests on VM

```
```


tdoehmen:
I wonder what was the initial problem here. Was it that the conversion didn't work for string-based date columns?
I'm afraid using the `convert_event_time_to_timestamp` function would lead to inconstistencies between python and java (see https://github.com/logicalclocks/feature-store-api/blob/master/java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java#L309)

python/hsfs/engine/spark.py
```diff
@@ -490,13 +491,17 @@ def _random_split(self, dataset, training_dataset):
...
     ):
+        # registering the UDF
+        _convert_event_time_to_timestamp = udf(
+            util.convert_event_time_to_timestamp, LongType()
+        )
+
         result_dfs = {}
         ts_type = dataset.select(event_time).dtypes[0][1]
         ts_col = (
             unix_timestamp(col(event_time)) * 1000
```
kennethmhc, line 502: the udf should be able to handle `"date", "timestamp"` as well, so we do not need the if-clause.

kennethmhc (changes requested):

kennethmhc:
> I wonder what was the initial problem here. Was it that the conversion didn't work for string-based date columns? I'm afraid using the `convert_event_time_to_timestamp` function would lead to inconstistencies between python and java (see https://github.com/logicalclocks/feature-store-api/blob/master/java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java#L309)

The problem was that we expect event time column has precision to second and in the backend we will multiple the value by 1000 for comparison. However, Davit use a column which is in millisecond, so the comparison did not work properly.

@davitbzh Like Till said, we should handle millisecond value in Java client as well.

python/hsfs/core/feature_view_engine.py
```diff
@@ -339,7 +339,7 @@ def _set_event_time(self, feature_view_obj, training_dataset_obj):
 
     def _get_start_time(self):
         # minimum start time is 1 second
-        return 1000
```
kennethmhc, line 342: Is this change necessary?

python/hsfs/engine/spark.py
```diff
@@ -492,14 +492,17 @@ def _random_split(self, dataset, training_dataset):
...
         self, training_dataset, dataset, event_time, drop_event_time=False
     ):
+        def _check_event_time_type(event_time):
+            # for backward compatibility
+            if isinstance(event_time, int) and len(str(event_time)) == 13:
+                event_time = int(event_time / 1000)
+            return util.convert_event_time_to_timestamp(event_time)
+
+        # registering the UDF
+        _convert_event_time_to_timestamp = udf(_check_event_time_type, LongType())
```
kennethmhc, line 502: can use `util.convert_event_time_to_timestamp` directly here.

java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java
```diff
@@ -303,16 +305,31 @@ private Dataset<Row>[] timeSeriesSplit(TrainingDataset trainingDataset, Query qu
...
+              .udf()
+              .register("checkEpochUDF", (Long input) -> {
+                if (Long.toString(input).length() == 13) {
+                  input = input / 1000;
+                  return input.longValue();
+                } else {
+                  return input;
+                }
+              }, DataTypes.LongType);
+          dataset = dataset.withColumn(eventTime,functions.callUDF(
```
kennethmhc, line 323: We should use a different name as the new column an drop it later, other this will change the value of the original event time column.

kennethmhc (changes requested):

python/hsfs/engine/spark.py
davitbzh: Well I don't want to put this part in `util.convert_event_time_to_timestamp`

```
            # for backward compatibility
            if isinstance(event_time, int) and len(str(event_time)) == 13:
                event_time = int(event_time / 1000)
```

python/hsfs/core/feature_view_engine.py
davitbzh: If we eventually will support only 10 digit timestamp yes

kennethmhc: But now we should any timestamp in second right?

python/hsfs/engine/spark.py
kennethmhc: in util.py, we have
```        
    elif isinstance(event_time, int):
        if event_time == 0:
            raise ValueError("Event time should be greater than 0.")
        # jdbc supports timestamp precision up to second only.
        if len(str(event_time)) < 13:
            event_time = event_time * 1000
        return event_time
```
so, below is redundant.
```
if isinstance(event_time, int) and len(str(event_time)) == 13:
                event_time = int(event_time / 1000)
```

davitbzh: honestly this is wrong:
```
        if len(str(event_time)) < 13:
            event_time = event_time * 1000
```

but I don't know how to replace without braking API. 

If we don't use this then how do we guarantee that even_time will have seconds granularity?

```
if isinstance(event_time, int) and len(str(event_time)) == 13:
                event_time = int(event_time / 1000)
```

davitbzh: I can check either 10 or 13?

java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java
```diff
@@ -310,6 +310,7 @@ private Dataset<Row>[] timeSeriesSplit(TrainingDataset trainingDataset, Query qu
             query.getLeftFeatureGroup().getFeature(eventTime).getType();
 
         if (BIGINT.getType().equals(eventTimeType)) {
+          String tmpEventTime = eventTime + "_tmp";
```
kennethmhc, line 313:
```suggestion
          String tmpEventTime = eventTime + "_hopsworks_tmp";
```

kennethmhc (approved):
LGTM！

java/src/main/java/com/logicalclocks/hsfs/engine/SparkEngine.java
```diff
@@ -303,27 +305,43 @@ private Dataset<Row>[] timeSeriesSplit(TrainingDataset trainingDataset, Query qu
...
         String eventTimeType =
-            query.getLeftFeatureGroup().getFeature(query.getLeftFeatureGroup().getEventTime()).getType();
+            query.getLeftFeatureGroup().getFeature(eventTime).getType();
+
         if (BIGINT.getType().equals(eventTimeType)) {
+          String tmpEventTime = eventTime + "_hopsworks_tmp";
+          sparkSession.sqlContext()
+              .udf()
+              .register("checkEpochUDF", (Long input) -> {
+                if (Long.toString(input).length() == 13) {
```
kennethmhc, line 317:
```suggestion
                if (Long.toString(input).length() > 10) {
```
To be consistent with python client.

PR #857, opened 2022-11-08
[FSTORE-418] bug fix Write options have no effect with .insert method in Python engine

davitbzh:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Manual Tests on VM

```
```


python/hsfs/core/feature_group_engine.py
```diff
@@ -402,6 +403,19 @@ def _save_feature_group_metadata(
...
+            _spark_options = write_options.pop("spark", None)
+            _write_options = (
+                [{"name": k, "value": v} for k, v in write_options.items()]
+                if write_options
+                else None
+            )
+            feature_group._deltastreamer_jobconf = DeltaStreamerJobConf(
+                _write_options, _spark_options
+            )
+
```
moritzmeister, line 418: To be honest, I am not quite sure why we have the first if statement in the first place, because the `_options` only get sent to the backend for external feature groups. 

However to keep the order of statemens from before, as we had it in the save, we need to have the `feature_group._options = write_options` after the `pop` call. If I am not mistaken, the pop modifies the dictionary in place, so in the old version, when `write_options` get passed to the feature group engine, it was already missing the spark key.

```suggestion
        if feature_group.stream:
            # when creating a stream feature group, users have the possibility of passing
            # a spark_job_configuration object as part of the write_options with the key "spark"
            _spark_options = write_options.pop("spark", None)
            feature_group._options = write_options
            _write_options = (
                [{"name": k, "value": v} for k, v in write_options.items()]
                if write_options
                else None
            )
            feature_group._deltastreamer_jobconf = DeltaStreamerJobConf(
                _write_options, _spark_options
            )

```

moritzmeister (changes requested):
Can we have a unit test for the `_save_feature_group_metadata` method. we can mock the api. But just to make sure the delta streamer job config is set correctly.

moritzmeister (approved):

PR #873, opened 2022-11-21
[FSTORE-476] Schema not updated for hsfs java client after append

bubriks:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/java/src/main/java/com/logicalclocks/hsfs/metadata/FeatureGroupBase.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/metadata/FeatureGroupBase.java
@@ -77,7 +77,6 @@ public class FeatureGroupBase {
   protected List<String> primaryKeys;
 
   @Getter
-  @Setter
   protected List<Feature> features;
 
   @Getter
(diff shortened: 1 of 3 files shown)


java/src/main/java/com/logicalclocks/hsfs/metadata/FeatureGroupBase.java
```diff
@@ -116,6 +115,11 @@ public FeatureGroupBase(FeatureStore featureStore, Integer id) {
     this.id = id;
   }
 
+  public void setFeatures(List<Feature> features) {
```
moritzmeister, line 118: Can we have at least a unit test so when someone goes in and doesn't know why we are doing this, doesn't remove it?

moritzmeister (changes requested):
I would prefer if we would handle this somewhere here in the updateFeatures and appendFeatures method https://github.com/logicalclocks/feature-store-api/blob/9f7628498b3a6bd0ee486062be89b040163b126f/java/src/main/java/com/logicalclocks/hsfs/engine/FeatureGroupBaseEngine.java#L95

moritzmeister (changes requested):
unit test please, you can mock the api call, just test that the subject is null after calling appendFeatures.

moritzmeister (approved):

PR #883, opened 2022-12-01
[FSTORE-309] Add support for like operator in query filter

bubriks:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/java/src/main/java/com/logicalclocks/hsfs/Feature.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/Feature.java
@@ -196,6 +196,10 @@ public class Feature {
     return new Filter(this, SqlFilterCondition.IN, jsonArray.toString());
   }
 
+  public Filter lk(Feature value) {
+    return new Filter(this, SqlFilterCondition.LIKE, value.toJson());
+  }
+
   public String toJson() {
     try {
       return mapper.writeValueAsString(this);
--- a/java/src/main/java/com/logicalclocks/hsfs/Feature.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/Feature.java
@@ -191,7 +191,15 @@ public class Feature {
     return new Filter(this, SqlFilterCondition.GREATER_THAN_OR_EQUAL, value.toJson());
   }
 
+  /**
+   * @deprecated
+   * `in` method is deprecated. Use `isin` instead.
+   */
   public Filter in(Collection<?> collection) {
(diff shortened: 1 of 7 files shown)


davitbzh (commented):
@bubriks is it possible to include code snippets for this, either here or in docs?

java/src/main/java/com/logicalclocks/hsfs/Feature.java
```diff
@@ -196,6 +196,10 @@ public Filter in(Collection<?> collection) {
     return new Filter(this, SqlFilterCondition.IN, jsonArray.toString());
   }
 
+  public Filter lk(Feature value) {
```
moritzmeister, line 199: Let's call it also `like` like in python?

moritzmeister (changes requested):

python/hsfs/feature.py
```diff
@@ -215,6 +215,9 @@ def __gt__(self, other):
     def contains(self, other):
         return filter.Filter(self, filter.Filter.IN, json.dumps(other))
 
+    def like(self, other):
```
moritzmeister, line 218: Should be possible to have a simple unit test for this?

java/src/main/java/com/logicalclocks/hsfs/Feature.java
bubriks: Ok will change. I saw all methods were 2 char so I also made this like that.

moritzmeister: yeah but prefer conistency over short cause we don't have java docs

bubriks: should i also rename "in" for "contains" (in python it is "contains")?

moritzmeister (approved):

java/src/main/java/com/logicalclocks/hsfs/Feature.java
moritzmeister: no I actually think "contains" is bad, `x contains ["avx"]` that sounds incorrect, or? the condition is actually `x in ["a", "b"]

moritzmeister (approved):

PR #885, opened 2022-12-05
[FSTORE-467] Make it easier to start the backfill job from python

bubriks:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/core/job.py
+++ b/python/hsfs/core/job.py
@@ -16,6 +16,7 @@
 
 import humps
 from hsfs import engine
+from hsfs.core import job_api
 
 
 class Job:
@@ -39,6 +40,8 @@ class Job:
         self._executions = executions
         self._href = href
 
+        self._job_api = job_api.JobApi()
+
     @classmethod
     def from_response_json(cls, json_dict):
         json_decamelized = humps.decamelize(json_dict)
@@ -60,5 +63,12 @@ class Job:
     def href(self):
         return self._href
 
-    def run(self, await_termination=True):
-        engine.get_instance().run_job(self, await_termination)
+    def run(self, write_options=None):
+        print(f"Launching job: {self.name}")
+        self._job_api.launch(self.name)
+        print(
+            "Job started successfully, you can follow the progress at \n{}".format(
+                engine.get_instance().get_job_url(self.href)
+            )
+        )
+        engine.get_instance().wait_for_job(self, write_options=write_options)
--- a/python/hsfs/core/job.py
+++ b/python/hsfs/core/job.py
@@ -63,7 +63,7 @@ class Job:
     def href(self):
         return self._href
 
-    def run(self, await_termination=True):
+    def run(self, await_termination: bool = True):
         """Run the job.
 
         Runs the job, by default awaiting its completion.
@@ -84,7 +84,7 @@ class Job:
             ```
 
         # Arguments
-            await_termination: bool. Identifies if the client should wait for the job to complete, defaults to True.
+            await_termination: Identifies if the client should wait for the job to complete, defaults to True.
         """
         print(f"Launching job: {self.name}")
         self._job_api.launch(self.name)
--- a/python/hsfs/core/job.py
+++ b/python/hsfs/core/job.py
@@ -15,6 +15,7 @@
 #
 
 import humps
+from hsfs import engine
 
 
 class Job:
@@ -58,3 +59,6 @@ class Job:
     @property
     def href(self):
         return self._href
+
+    def run(self, write_options=None):
+        engine.get_instance().run_job(self, write_options)
--- a/python/hsfs/core/job.py
+++ b/python/hsfs/core/job.py
@@ -64,6 +64,28 @@ class Job:
         return self._href
 
     def run(self, await_termination=True):
+        """Run the job.
+
+        Runs the job, by default awaiting its completion.
+
+        !!! example
+            ```python
+            # connect to the Feature Store
+            fs = ...
+
+            # get the Feature Group instances
+            fg = fs.get_or_create_feature_group(...)
+
+            # insert in to feature group
+            job, _ = fg.insert(df, write_options={"start_offline_backfill": False})
+
+            # run job
+            job.run()
+            ```
+
+        # Arguments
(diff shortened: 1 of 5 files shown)


python/hsfs/core/job.py
```diff
@@ -58,3 +59,6 @@ def executions(self):
     @property
     def href(self):
         return self._href
+
+    def run(self, write_options=None):
```
moritzmeister, line 63: We shouldn't leak "feature store concepts" like `write_options` to the jobs api.

python/hsfs/engine/python.py
```diff
@@ -933,17 +925,20 @@ def acked(err, msg):
...
 
+    def run_job(self, job, write_options=None):
+        print(f"Launching job: {job.name}")
+        self._job_api.launch(job.name)
+        print(
+            "Job started successfully, you can follow the progress at \n{}".format(
+                self._get_job_url(job.href)
+            )
+        )
+        self._wait_for_job(job, write_options)
```
moritzmeister, line 940: I would move this method to the Job class and replace the `write_options` with a simple parameter `await_termination` so that you don't need to leak the feature store specific write_options to the jobs api.

You might need to move the `JobsApi` class to the `Jobs` class, but I think that's ok.

python/hsfs/engine/python.py
```diff
@@ -933,17 +925,20 @@ def acked(err, msg):
...
         ):
-            print("Launching offline feature group backfill job...")
-            self._job_api.launch(job_name)
-            print(
-                "Backfill Job started successfully, you can follow the progress at \n{}".format(
-                    self._get_job_url(job.href)
-                )
-            )
-            self._wait_for_job(job, offline_write_options)
+            self.run_job(job, offline_write_options)
```
moritzmeister, line 928: If you do what I suggest in the other comments, you should be able to replace this with

```python
job.run(await_termination=write_options.get("wait_for_job", True))
```

python/hsfs/engine/python.py
bubriks: there is the following issue:
`tests/engine/test_python.py:None (tests/engine/test_python.py)
test_python.py:23: in <module>
    from hsfs import (
../../hsfs/__init__.py:20: in <module>
    from hsfs.connection import Connection
../../hsfs/connection.py:23: in <module>
    from hsfs import engine, client, util
../../hsfs/engine/__init__.py:17: in <module>
    from hsfs.engine import spark
../../hsfs/engine/spark.py:67: in <module>
    from hsfs import feature, training_dataset_feature, client, util
../../hsfs/training_dataset_feature.py:19: in <module>
    import hsfs.feature_group as feature_group
../../hsfs/feature_group.py:31: in <module>
    from hsfs.core import (
../../hsfs/core/feature_group_engine.py:21: in <module>
    from hsfs.core import feature_group_base_engine, hudi_engine
../../hsfs/core/feature_group_base_engine.py:17: in <module>
    from hsfs.core import feature_group_api, storage_connector_api, tags_api, kafka_api
../../hsfs/core/feature_group_api.py:19: in <module>
    from hsfs.core import ingestion_job
../../hsfs/core/ingestion_job.py:18: in <module>
    from hsfs.core.job import Job
../../hsfs/core/job.py:19: in <module>
    from hsfs.core import job_api
../../hsfs/core/job_api.py:21: in <module>
    class JobApi:
../../hsfs/core/job_api.py:39: in JobApi
    def get(self, name: str) -> job.Job:
E   AttributeError: partially initialized module 'hsfs.core.job' has no attribute 'Job' (most likely due to a circular import)
`

python/hsfs/engine/python.py
```diff
@@ -778,12 +770,10 @@ def _get_app_options(self, user_write_options={}):
...
         )
 
-    def _wait_for_job(self, job, user_write_options=None):
+    def _wait_for_job(self, job, await_termination=True):
         # If the user passed the wait_for_job option consider it,
         # otherwise use the default True
-        while user_write_options is None or user_write_options.get(
-            "wait_for_job", True
-        ):
+        while await_termination:
```
moritzmeister, line 776: Ok, if we have to do it like this, one more thing how to make the code a bit cleaner.
Leave this method as it is and just change it in the `Job.run()` by simply wrapping the parameter in a dict? so we don't have to have the if statement spread through the code?

python/hsfs/core/job.py
```diff
@@ -58,3 +59,6 @@ def executions(self):
     @property
     def href(self):
         return self._href
+
+    def run(self, await_termination=True):
+        engine.get_instance().run_job(self, await_termination)
```
moritzmeister, line 64: see other comment, one more suggestion
```suggestion
        engine.get_instance().run_job(self, {"wait_for_job": await_termination})
```

python/hsfs/engine/python.py
bubriks: i could do it, but it impacts only 1 place, while run is called in 2 places (where we have to do the check)

python/hsfs/core/job.py
```diff
@@ -58,3 +62,35 @@ def executions(self):
...
+
+            # insert in to feature group
+            job, _ = fg.insert(df, write_options={"start_offline_backfill": False})
+
+            # run job
+            job.run()
+            ```
+
+        # Arguments
+            await_termination: bool. Identifies if the client should wait for the job to complete, defaults to True.
```
moritzmeister, line 87: if you add the type annotation, you don't need to write it in text in the docstring
```suggestion
    def run(self, await_termination: bool = True):
        """Run the job.

        Runs the job, by default awaiting its completion.

        !!! example
            ```python
            # connect to the Feature Store
            fs = ...

            # get the Feature Group instances
            fg = fs.get_or_create_feature_group(...)

            # insert in to feature group
            job, _ = fg.insert(df, write_options={"start_offline_backfill": False})

            # run job
            job.run()
            ```

        # Arguments
            await_termination: Identifies if the client should wait for the job to complete, defaults to True.
```

moritzmeister (approved):

PR #925, opened 2023-02-08
[FSTORE-607] Investigate hsfs java resource leaks

bubriks:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/java/src/main/java/com/logicalclocks/hsfs/ExternalFeatureGroup.java
+++ b/java/src/main/java/com/logicalclocks/hsfs/ExternalFeatureGroup.java
@@ -58,10 +58,6 @@ public class ExternalFeatureGroup extends FeatureGroupBase {
   @Setter
   private List<OnDemandOptions> options;
 
-  @Getter
-  @Setter
-  private String type = "onDemandFeaturegroupDTO";
-
   private ExternalFeatureGroupEngine externalFeatureGroupEngine = new ExternalFeatureGroupEngine();
   private final CodeEngine codeEngine = new CodeEngine(EntityEndpointType.FEATURE_GROUP);
 
@@ -70,6 +66,7 @@ public class ExternalFeatureGroup extends FeatureGroupBase {
                               ExternalDataFormat dataFormat, String path, Map<String, String> options,
                               @NonNull StorageConnector storageConnector, String description, List<String> primaryKeys,
                               List<Feature> features, StatisticsConfig statisticsConfig, String eventTime) {
+    this();
     this.timeTravelFormat = null;
     this.featureStore = featureStore;
     this.name = name;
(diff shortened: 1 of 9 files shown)


java/src/main/java/com/logicalclocks/hsfs/ExternalFeatureGroup.java
```diff
@@ -70,6 +66,7 @@ public ExternalFeatureGroup(FeatureStore featureStore, @NonNull String name, Int
                               ExternalDataFormat dataFormat, String path, Map<String, String> options,
                               @NonNull StorageConnector storageConnector, String description, List<String> primaryKeys,
                               List<Feature> features, StatisticsConfig statisticsConfig, String eventTime) {
+    this();
```
moritzmeister, line 69: Don't quite understand why this is necessary?

bubriks: this calls the constructor without any parameters it assigns the fg type https://github.com/logicalclocks/feature-store-api/pull/925/files/28bbaa23949fb500be7f85f7167fa501c484dc29#diff-ee497c159fe67d2d377c07b9ab8644c83eafc35987124649ccac254ecc424653R90 (it works like this for all types of fg, since otherwise even though we had type defined in FeatureGroupBase we always overwrote it)

moritzmeister: sure I understand how it works. But it seems error prone, whenever someone adds a constructor, we need to make sure we don't forget calling `this()`.
What was the problem with having it set for the class?

bubriks: true i agree, it exposes us to more developer errors, but at the same time, we define the same variable in both the base class and the class that extends the base class. 

This is the exact issue I was trying to resolve: https://sonarqube.hops.works/project/issues?id=hsfs-java&open=AYXOxaSNyM9k4vySez5d&resolved=false&severities=BLOCKER

I think there might be 2 ways to enforce that all necessary variables are assigned.
1: make "getType"/"setType" method into an abstract method and force each sub-class implementing the super-class to define the value it should return.
2. Do we even need this variable "type"? maybe we could remove it and resolve the issue that way.

Another thing I think we could do to help developers from having issues is to make "FeatureGroupBase" into abstract class (so that no objects could be created of it directly). `public class FeatureGroupBase {` -> `public abstract class FeatureGroupBase {`

moritzmeister: yeah that's right, an abstract class is the right approach.
I think the backend needs the type parameter.

moritzmeister: Anyway, I will accept this for now.

PR #927, opened 2023-02-13
[FSTORE-680] Optimize HSFS for single row writes

moritzmeister:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/core/feature_group_engine.py
+++ b/python/hsfs/core/feature_group_engine.py
@@ -58,7 +58,7 @@ class FeatureGroupEngine(feature_group_base_engine.FeatureGroupBaseEngine):
             return None, ge_report
 
         offline_write_options = write_options
-        online_write_options = self.get_kafka_config(write_options)
+        online_write_options = write_options
 
         return (
             engine.get_instance().save_dataframe(
@@ -114,7 +114,7 @@ class FeatureGroupEngine(feature_group_base_engine.FeatureGroupBaseEngine):
             return None, ge_report
 
         offline_write_options = write_options
-        online_write_options = self.get_kafka_config(write_options)
+        online_write_options = write_options
 
         if not feature_group.online_enabled and storage == "online":
             raise exceptions.FeatureStoreException(
@@ -238,26 +238,6 @@ class FeatureGroupEngine(feature_group_base_engine.FeatureGroupBaseEngine):
     def get_subject(self, feature_group):
         return self._kafka_api.get_topic_subject(feature_group._online_topic_name)
 
-    def get_kafka_config(self, online_write_options):
-        # This should be in the Spark engine
-        if self._kafka_config is None:
-            self._kafka_config = {
-                "kafka.bootstrap.servers": ",".join(
-                    [
-                        endpoint.replace("INTERNAL://", "")
-                        for endpoint in self._kafka_api.get_broker_endpoints()
-                    ]
-                ),
-                "kafka.security.protocol": "SSL",
-                "kafka.ssl.truststore.location": client.get_instance()._get_jks_trust_store_path(),
-                "kafka.ssl.truststore.password": client.get_instance()._cert_key,
-                "kafka.ssl.keystore.location": client.get_instance()._get_jks_key_store_path(),
-                "kafka.ssl.keystore.password": client.get_instance()._cert_key,
-                "kafka.ssl.key.password": client.get_instance()._cert_key,
-                "kafka.ssl.endpoint.identification.algorithm": "",
-            }
-        return {**online_write_options, **self._kafka_config}
-
     def insert_stream(
         self,
         feature_group,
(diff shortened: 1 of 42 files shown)


python/hsfs/engine/python.py
```diff
@@ -819,16 +848,9 @@ def acked(err, msg):
...
-        # initialize progress bar
-        progress_bar = tqdm(
-            total=dataframe.shape[0],
-            bar_format="{desc}: {percentage:.2f}% |{bar}| Rows {n_fmt}/{total_fmt} | "
-            "Elapsed Time: {elapsed} | Remaining Time: {remaining}",
-            desc="Uploading Dataframe",
-            mininterval=1,
-        )
+            if not feature_group._multi_part_insert:
+                progress_bar.update()
```
bubriks, line 852: is the user somehow informed of the progress being made for multi-insert?

bubriks: maybe we could write how many entries have been submitted.

moritzmeister: I guess the right approach would be to keep the progress bar similar to the producer in the feature group object.
But I am not sure how it will look like then and if it works to update it across function calls. This is an edge case mostly done in applications and high performance, not interactively in notebooks, so my thinking was that the progress bar is not important and will just slow it down.

python/hsfs/core/feature_group_engine.py
```diff
@@ -57,7 +57,7 @@ def save(
             return None, ge_report
 
         offline_write_options = write_options
-        online_write_options = self.get_kafka_config(write_options)
+        online_write_options = write_options
```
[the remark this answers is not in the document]
moritzmeister: yeah for the moment I left it because I was afraid some downstream calls make modifications to the original dict. however, I think since we don't make a copy, the assignment to a new variable won't help in that case either.

moritzmeister:
@bubriks @SirOibaf ok I think I am done with unit tests, if you want to review

bubriks (approved):

python/hsfs/engine/python.py
```diff
@@ -807,6 +803,39 @@ def _write_dataframe_kafka(
...
+                producer = feature_group._kafka_producer
+                feature_writers = feature_group._feature_writers
+                writer = feature_group._writer
+        else:
+            producer, feature_writers, writer = self._init_kafka_resources(
+                feature_group, offline_write_options
+            )
+
+            # initialize progress bar
+            progress_bar = tqdm(
```
SirOibaf, line 832: Why is the progress bar initialized only in the else statement?

SirOibaf: Nevermind seen the comment below.

python/hsfs/feature_group.py
```diff
@@ -1675,6 +1698,151 @@ def insert(
...
+            TypeVar("pyspark.RDD"),  # noqa: F821
+            np.ndarray,
+            List[list],
+        ] = None,
+        overwrite: Optional[bool] = False,
+        operation: Optional[str] = "upsert",
+        storage: Optional[str] = None,
+        write_options: Optional[Dict[str, Any]] = {},
+        validation_options: Optional[Dict[str, Any]] = {},
+    ) -> Tuple[Optional[Job], Optional[ValidationReport]]:
```
SirOibaf, line 1715: Shouldn't this include the writer as well?

python/hsfs/feature_group.py
```diff
@@ -1675,6 +1698,151 @@ def insert(
...
+                When using the `python` engine, write_options can contain the
+                following entries:
+                * key `spark` and value an object of type
+                [hsfs.core.job_configuration.JobConfiguration](../job_configuration)
+                  to configure the Hopsworks Job used to write data into the
+                  feature group.
+                * key `wait_for_job` and value `True` or `False` to configure
+                  whether or not to the insert call should return only
+                  after the Hopsworks Job has finished. By default it waits.
+                * key `start_offline_backfill` and value `True` or `False` to configure
```
SirOibaf, line 1783: `start_offline_backfill` is False by default in multi part insert.

python/hsfs/feature_group.py
```diff
@@ -1675,6 +1698,151 @@ def insert(
...
+
+        For these cases, the feature group provides the `multi_part_insert` API,
+        which is optimized for writing many small Dataframes after another.
+
+        There are two ways to use this API:
+        !!! example "Python Context Manager"
+            Using the Python `with` syntax you can acquire a FeatureGroupWriter
+            object that implements the same `multi_part_insert` API.
+            ```python
+            feature_group = fs.get_or_create_feature_group("fg_name", version=1)
```
SirOibaf, line 1732: Is there an assumption that users call `save()` before using the `multi_part_insert()` method?

SirOibaf (changes requested):

python/hsfs/feature_group.py
moritzmeister: changed, plus added comment above on how to start backfill job once the multi part insert is done.

python/hsfs/feature_group.py
moritzmeister: true, done

SirOibaf (approved):

PR #933, opened 2023-02-16
[FSTORE-693] Fix validation_time with timezone

vatj:
This PR adds/fixes/changes...
Add dateutil parsing for validation_time provided as str to support timezone

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/ge_validation_result.py
+++ b/python/hsfs/ge_validation_result.py
@@ -16,7 +16,8 @@
 
 import json
 from typing import Any, Dict, Optional, Union
-from datetime import date, datetime
+import datetime
+import dateutil
 
 import humps
 import great_expectations as ge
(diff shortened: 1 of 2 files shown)


python/hsfs/ge_validation_result.py
```diff
@@ -199,6 +202,13 @@ def validation_time(
...
+        if isinstance(validation_time, str):
+            try:
+                validation_time = dateutil.parser.parse(validation_time).astimezone(
+                    datetime.timezone.utc
+                )
+            except ValueError:
+                pass
         if validation_time:
             self._validation_time = util.convert_event_time_to_timestamp(
                 validation_time
```
SirOibaf, line 214: I'm confused here. Why do we need this second if statement? Is there a case where the user sets the validation_time manually?

vatj: I think for consistency with other places where we are dealing with start_time/end_time it is good to use the same function

vatj: And removing it would loose compatibility if any user has been doing something where the time is set with something else than a str

SirOibaf: Ok fair enough. Maybe can you add a comment on why the second if is needed?

SirOibaf (approved):

PR #948, opened 2023-03-09
[FSTORE-690] Setup Online Feature Store Benchmark with Locust

moritzmeister:
- don't send producer config to backend
- filter
- lookup working
- stuff
- readme

This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


locust_benchmark/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM hopsworks/locust:hsfs-latest
+
+USER root
+
+RUN apt-get update -y
+RUN apt-get -y install gcc python3-dev librdkafka-dev git
```
olapiv, line 6: Run as single command otherwise `update` will always be cached and the installs will run on outdated apt infos if the command is changed.

moritzmeister: fixed, thanks

locust_benchmark/README.md
```diff
@@ -0,0 +1,178 @@
...
+      cd locust_benchmark
+      pip install -r requirements.txt
+      ```
+
+#### Creating feature group for lookups
+
+1. Save the previously created API key in a file named `.api_key`
+
+```bash
+echo "[YOUR KEY]" > .api_key
```
olapiv, line 64: Perhaps just add an env file, with all configs, which you source? Or just write it into the config?

locust_benchmark/README.md
```diff
@@ -0,0 +1,178 @@
...
+```bash
+curl -fsSL https://get.docker.com -o get-docker.sh
+sudo sh get-docker.sh
+```
+
+#### Load the docker image
+
+```bash
+wget https://repo.hops.works/dev/moritz/locust_benchmark/locust-2.15.0-hsfs-3.2.0.dev1-amd64.tgz
+sudo docker load < locust-2.15.0-hsfs-3.2.0.dev1-amd64.tgz
```
olapiv, line 141: Why not just use www.dockerhub.com/u/hopsworks? Just refer to `hopsworks/locust:2.15` in the docker-compose file, nothing needs to be pulled/downloaded manually then.

olapiv (approved):
Looks good. I'd personally just cut all the checkout master, single-threaded, install Anaconda, etc. stuff and just say "Dependency: Docker" and use an image that you've uploaded to Dockerhub. I also don't see the point in explaining how to install Docker. Gives more space to explain how to interpret Locust results in the README.

For development use, you can still run `docker build` first and then reference your local image in the docker-compose file.

locust_benchmark/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
+version: '3'
+
+services:
+  master:
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
```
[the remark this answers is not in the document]
moritzmeister: if you run it against a dev VM where you need to connect to hopsworks on localhost, I needed this to be able to reach the ports inside the containers. Bit of an edge case, as it's just us who will run against dev VMs. could remove it, but it also doesn't hurt.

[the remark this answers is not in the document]
olapiv: Okay, can add a comment `just in case we're running on the same machine as Hopsworks`

PR #963, opened 2023-03-22
[FSTORE-771] UX proposal fluent API feature monitoring

vatj:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


python/hsfs/core/feature_descriptive_statistics.py
```diff
@@ -244,3 +244,31 @@ def exact_num_distinct_values(self) -> Optional[int]:
...
+    @property
+    def row_percentage(self) -> float:
+        return self._row_percentage
+
+    @row_percentage.setter
+    def row_percentage(self, row_percentage: float):
+        if isinstance(row_percentage, int) or isinstance(row_percentage, float):
+            row_percentage = float(row_percentage)
+            if row_percentage <= 0.0 or row_percentage > 1.0:
+                raise ValueError("Row percentage must be a float between 0 and 1.")
```
javierdlrm, line 269: So we don't support number of rows?

python/hsfs/core/feature_monitoring_config.py
```diff
@@ -87,16 +108,16 @@ def from_response_json(cls, json_dict):
...
+    # def _parse_schedule_config(
+    #     self, schedule_config: Optional[Union[JobScheduler, dict]]
+    # ):
+    #     if schedule_config is None:
+    #         return None
+    #     return (
+    #         schedule_config
+    #         if isinstance(schedule_config, JobScheduler)
+    #         else JobScheduler.from_response_json(schedule_config)
+    #     )
```
javierdlrm, line 121: leftover?

python/hsfs/core/feature_descriptive_statistics.py
vatj: Spark does not via .sample so we will stick to percentage. Maybe change the name to row_fraction, what do you think?

python/hsfs/feature_group.py
```diff
@@ -1078,6 +1094,281 @@ def validate(
             ge_type=ge_type,
         )
 
+    def _get_feature_monitoring_configs(
```
kennethmhc, line 1097: Are the new methods intended to be accessible by users? If so, why are they prefixed by `_`?

python/hsfs/feature_group.py
```diff
@@ -1078,6 +1094,281 @@ def validate(
...
+
+        # Return
+            `FeatureMonitoringConfig` Configuration with minimal information about the feature monitoring.
+                Additional information are required before feature monitoring is enabled.
+        """
+        if not self._id:
+            raise FeatureStoreException(
+                "Only Feature Group registered with Hopsworks can enable scheduled statistics monitoring."
+            )
+        fm_engine = feature_monitoring_config_engine.FeatureMonitoringConfigEngine(
```
kennethmhc, line 1286: can reuse `self._feature_monitoring_config_engine`?

python/hsfs/feature_group.py
```diff
@@ -1078,6 +1094,281 @@ def validate(
...
+            feature_store_id=self._feature_store_id,
+            feature_group_id=self._id,
+        )
+
+        return fm_engine._build_default_scheduled_statistics_config(
```
kennethmhc, line 1291: The config did not send to the backend.

kennethmhc (commented):
A few questions

python/hsfs/core/feature_descriptive_statistics.py
DhananjayMukhedkar: For me atleast percentage is clearer, as fraction could be based on any number/float  but percentage is based on 100 or 1 in this case.

python/hsfs/feature_group.py
DhananjayMukhedkar, line 1097: General question, why are these APIs private ?  should'nt it be a public without the underscore if its client facing?

DhananjayMukhedkar:
can you check the auto_doc failure while docs creation?

python/hsfs/feature_group.py
vatj: Because we want to be able to merge in master but not make the UX public while we are stabilising it

vatj: See kenneth comment

python/hsfs/feature_group.py
vatj: For now this is intended. Because we want to restrict the possible updates to the config to allow the history to be meaningful, we cannot just register it directly and then update it with a modified detection window. So they get a config back and need to call .save() to register it once they are done setting it up. Needs to be made clearer in docstring maybe

kennethmhc: yea just looking at the method name and the description, users do not know that they need to call `.save()` to commit the config.

python/hsfs/feature_view.py
```diff
@@ -2128,6 +2124,303 @@ def delete_all_training_datasets(self):
...
+
+        return fm_result_engine.get_feature_monitoring_results(
+            config_name=config_name,
+            config_id=config_id,
+            start_time=start_time,
+            end_time=end_time,
+            with_statistics=with_statistics,
+        )
+
+    def _enable_scheduled_statistics_monitoring_fluent(
```
kennethmhc, line 2277: The method name does not match what it really does. The purpose of the method is `create_fm_config`.

python/hsfs/feature_view.py
```diff
@@ -2128,6 +2124,303 @@ def delete_all_training_datasets(self):
...
+            )
+        )
+
+        return fm_config_engine.get_feature_monitoring_configs(
+            name=name,
+            feature_name=feature_name,
+            config_id=config_id,
+        )
+
+    def _get_feature_monitoring_history(
```
kennethmhc, line 2201: This is a duplicate of `feature_monitoring_config.get_history`. Instead users can do `fv.get_config.get_history`

python/hsfs/feature_view.py
```diff
@@ -2128,6 +2124,303 @@ def delete_all_training_datasets(self):
         """
         self._feature_view_engine.delete_training_data(self)
 
+    def _get_feature_monitoring_configs(
```
kennethmhc, line 2127: should have a `get_scheduled_statistics_monitoring_configs` also? users may not know they the config id are shared between statistics config and fm config.

vatj: Ja that is one of the question. Maybe we could rename it to monitoring config to implicitly cover both cases?

python/hsfs/feature_view.py
vatj: I realise that, I do feel like having to go through the config is less practical for the user because you have to get the config before being able to either:
- autocomplete with a method that does what you want
- look at the docs for the get_history function

python/hsfs/feature_view.py
vatj: We could rename both methods from enable to create

PR #969, opened 2023-03-24
Feature monitoring API at feature group level dict proposal

javierdlrm:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/core/monitoring_window_config.py
+++ b/python/hsfs/core/monitoring_window_config.py
@@ -57,7 +57,7 @@ class WindowConfigType(str, Enum):
 
 
 class MonitoringWindowConfig:
-    _DEFAULT_ROW_PERCENTAGE = 1.0
+    DEFAULT_ROW_PERCENTAGE = 1.0
 
     def __init__(
         self,
@@ -274,7 +274,7 @@ class MonitoringWindowConfig:
         self._specific_value = specific_value
 
     @property
-    def row_percentage(self) -> Optional[int]:
+    def row_percentage(self) -> Optional[float]:
         return self._row_percentage
 
     @row_percentage.setter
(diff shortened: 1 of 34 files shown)


python/hsfs/feature_group.py
```diff
@@ -974,6 +979,196 @@ def get_validation_history(
...
+            with_reference_window: Configuration of the reference window.
+            compare_on: Configuration of the statistics comparison.
+            alert_on: Configuration of the alert.
+            on_schedule: Configuration of the scheduler.
+            description: Description of the monitoring configuration.
+
+        # Returns:
+            `FeatureMonitoringConfig` Feature monitoring configuration
+        """
+        fm_config_engine = self._get_instance_feature_monitoring_config_engine()
```
vatj, line 1149: You can skip assigning it to a variable

python/hsfs/core/monitoring_window_config.py
```diff
@@ -67,7 +67,7 @@ def __str__(self):
         return self.json()
 
     def __repr__(self):
-        return json.dumps(humps.decamelize(self.to_dict()), indent=2)
+        return f"MonitoringWindowConfig(type: {self._window_config_type})"
```
vatj, line 70: I don't think we should make it harder to introspect entities quickly, even for us when we need to do debugging

vatj (commented):
I think you should put some more thinking in good defaults if we want to go forward with this UX. For the sake of clarity can you add a comment to method which are intended to be public when it stabilises

python/hsfs/core/monitoring_window_config.py
javierdlrm: For debugging purposes, I agree. But for introspection I think it's better to have a specific .describe() method.

PR #973, opened 2023-03-28
[FSTORE-744] Flying Duck Integration

tdoehmen:
This PR adds/fixes/changes...
- add read_options={"use_hive": True/False} to feature_group/query.read(, read_options) and featureview.trainining_data(..read_options)/train_test_split(..read_options)/train_test_validation_split(..read_options) and featureview.get_batch_data()
- calls flyingduck when feature flag for flyingduck is enabled and connection could be established
- does not contain: create_trainning_data/create_train_test_split/create_train_test_validation_split integration with flyingduck

JIRA Issue: FSTORE-744

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


tdoehmen:
Still needed:
- a variable for the flyingduck ip/port in hopsworks 
- disable flyingduck if vaiables are not available (e.g. in community)

python/hsfs/core/arrow_flight_client.py
```diff
@@ -0,0 +1,371 @@
...
+                )
+            filters = (filters & join_filters) if join_filters is not None else filters
+
+        return featuregroups, features, filters
+
+    def _filter_to_expression(self, filters, featuregroups):
+        if not filters:
+            return None
+        filter_expression = self._resolve_logic(filters, featuregroups)
+        return filter_expression
```
bubriks, line 326:
```suggestion
        return self._resolve_logic(filters, featuregroups)
```

python/hsfs/core/arrow_flight_client.py
```diff
@@ -0,0 +1,371 @@
...
+        if logic._left_f:
+            filter_expression["left_filter"] = self._resolve_filter(
+                logic._left_f, featuregroups
+            )
+        elif logic._left_l:
+            filter_expression["left_filter"] = self._resolve_logic(
+                logic._left_l, featuregroups
+            )
+        else:
+            filter_expression["left_filter"] = None
```
bubriks, line 341: to avoid duplication (there could be some misspells in the key value), if changed here should also be done for right filter. Furthermore, right and left filter is done the same way, maybe make a new method, to remove this duplication.
```suggestion
        if logic._left_f:
            left_filter = self._resolve_filter(
                logic._left_f, featuregroups
            )
        elif logic._left_l:
            left_filter = self._resolve_logic(
                logic._left_l, featuregroups
            )
        else:
            left_filter = None
        filter_expression["left_filter"] = left_filter
```

bubriks (changes requested):
some more comments

python/tests/engine/test_python.py
```diff
@@ -296,8 +296,8 @@ def test_read_pandas_other(self, mocker):
 
     def test_read_hopsfs(self, mocker):
         # Arrange
-        mock_python_engine_read_hopsfs_rest = mocker.patch(
```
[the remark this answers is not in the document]
tdoehmen: I agree, I am not sure entirely sure how to test them, but will think about it

python/hsfs/engine/python.py
```diff
@@ -161,12 +181,19 @@ def _read_pandas(self, data_format, obj):
...
-    def _read_hopsfs(self, location, data_format):
+    def _read_hopsfs(self, location, data_format, read_options={}):
         # providing more informative error
         try:
             from pydoop import hdfs
         except ModuleNotFoundError:
-            return self._read_hopsfs_rest(location, data_format)
+            if self._arrow_flight_client.is_data_format_supported(
+                data_format
+            ) and self._arrow_flight_client.should_be_used(read_options):
```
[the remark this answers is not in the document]
tdoehmen: I tried to simplify it a bit

python/hsfs/constructor/query.py
```diff
@@ -64,6 +64,12 @@ def _prep_read(self, online, read_options):
                 sql_query = query.query
             online_conn = None
 
+            if engine.get_instance().flyingduck_supported_and_enabled(
+                self, read_options
+            ):
+                read_options["use_flyingduck"] = True
+                return (self, sql_query), online_conn
+
```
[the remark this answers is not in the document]
tdoehmen: read_options["use_flyingduck"] is something I added initially only for internal use. I first check if the feature flag for flyingduck is set (arrow_flight_server.is_enabled()), and the check if the query is supported; if both conditions are met I set read_options["use_flyingduck"]=True so the python engine knows it should call flyingduck instead of hive. Only later I added a check to see if read_options["use_flyingduck"]  was already set by the user. The ise was that the use could overrule the feature flag and query flyignduck even if it is disabled (mainly for testing). But I reverted the changes now. Hope it's clearer now.

python/hsfs/core/arrow_flight_client.py
```diff
@@ -0,0 +1,371 @@
...
+                return True
+            except Exception as e:
+                warnings.warn(
+                    f"Could not establish connection to FlyingDuck. ({e}) "
+                    f"Will fall back to spark for this operation. "
+                    f"If the error persists, you can disable FlyingDuck for this operation "
+                    f"by changing the read_options (read_options={{'enable_flyingduck'=False}} "
+                    f"or read_options={{'use_spark'=True}})."
+                )
+                return False
```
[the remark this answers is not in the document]
tdoehmen: See above, I simplified the logic and ignore the "use_flyginduck" read-option set by the user. Whether flyingduck is used or not, only depends now on the feature flag in the cluster config, whether read_options["use_spark"] is explicitiyl set, and on wheter the query/data format is supported.

python/hsfs/core/arrow_flight_client.py
```diff
@@ -0,0 +1,371 @@
...
+    return _arrow_flight_instance
+
+
+class ArrowFlightClient:
+    def __init__(self):
+        self._is_initialized = False
+
+        try:
+            self._variable_api = VariableApi()
+            self._is_enabled = self._variable_api.get_flyingduck_enabled()
```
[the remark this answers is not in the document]
tdoehmen: I might need the variable API at a later point, e.g. to get the endpoint URL for external clients.

python/hsfs/core/arrow_flight_client.py
```diff
@@ -0,0 +1,371 @@
...
+    def _initialize_connection_if_needed(self):
+        if self._is_initialized is True:
+            return
+
+        self._client = client.get_instance()
+
+        host_url = "grpc+tls://flyingduck.service.consul:5005"
+        (tls_root_certs, cert_chain, private_key) = self._extract_certs(self._client)
+        self._connection = pyarrow.flight.FlightClient(
+            location=host_url,
```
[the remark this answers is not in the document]
tdoehmen: Same here, for external clients we will need some more logic around the host_url

python/hsfs/core/arrow_flight_client.py
```diff
@@ -0,0 +1,371 @@
...
+def get_instance():
+    global _arrow_flight_instance
+    if not _arrow_flight_instance:
+        _arrow_flight_instance = ArrowFlightClient()
+    return _arrow_flight_instance
+
+
+class ArrowFlightClient:
+    def __init__(self):
+        self._is_initialized = False
```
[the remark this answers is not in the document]
tdoehmen: There was an option to bypass the featureflag by setting read_option["use_flyingduck"]=True. In that case I needed to 1) know if the connection was already initialized 2) be able to initialize it if needed. Since I removed the option to bypass the feature flag, we could, theoretically remove this. However I'm thinking we might still need it in the future.

python/hsfs/core/arrow_flight_client.py
```diff
@@ -0,0 +1,335 @@
...
+            and read_options["use_spark"] is True
+        ):
+            return False
+
+        if self._is_enabled:
+            return True
+
+    def is_query_supported(self, query, read_options):
+        supported = self._is_query_supported_rec(query)
+        return supported and self.should_be_used(read_options)
```
bubriks, line 88:
```suggestion
        return self._is_query_supported_rec(query) and self.should_be_used(read_options)
```

python/hsfs/core/arrow_flight_client.py
```diff
@@ -0,0 +1,335 @@
...
+            filters,
+        ) = self._collect_featuregroups_features_and_filters(query)
+
+        query = {
+            "query_string": query_string,
+            "featuregroups": featuregroups,
+            "features": features,
+            "filters": filters,
+        }
+        return query
```
bubriks, line 236:
```suggestion
        training_dataset_metadata["query"] = self._get_query_object(
            feature_view.query
        )
        training_dataset_metadata[
            "featurestore_name"
        ] = feature_view.query._left_feature_group.feature_store_name.replace(
            "_featurestore", ""
        )

        return training_dataset_metadata

    def _get_query_object(self, query):
        query_string = query.to_string().replace(
            f"`{query._left_feature_group.feature_store_name}`.`",
            f"`{query._left_feature_group.feature_store_name.replace('_featurestore','')}.",
        ).replace("`", '"')

        (
            featuregroups,
            features,
            filters,
        ) = self._collect_featuregroups_features_and_filters(query)

        query = {
            "query_string": query_string,
            "featuregroups": featuregroups,
            "features": features,
            "filters": filters,
        }
        return query
```

python/hsfs/core/arrow_flight_client.py
```diff
@@ -0,0 +1,335 @@
...
+        ) = self._collect_featuregroups_features_and_filters_rec(query)
+        filters = self._filter_to_expression(filters, featuregroups)
+        for feature in features:
+            features[feature] = list(features[feature])
+        return featuregroups, features, filters
+
+    def _collect_featuregroups_features_and_filters_rec(self, query):
+        featuregroups = {}
+        fg = query._left_feature_group
+        fg_name = f"{fg.feature_store_name.replace('_featurestore','')}.{fg.name}_{fg.version}"  # featurestore.name_version
```
bubriks, line 257: instead of always doing this `fg.feature_store_name.replace('_featurestore','')` in multiple places. i think it would be better to add a new function to FeatureGoupBase and instead call `fg.project_name` (it'll do this action)

python/hsfs/core/arrow_flight_client.py
```diff
@@ -0,0 +1,335 @@
...
+        training_dataset_metadata["query"] = self._get_query_object(
+            feature_view.query, feature_view.query.to_string()
+        )
+        training_dataset_metadata[
+            "featurestore_name"
+        ] = feature_view.query._left_feature_group.feature_store_name.replace(
+            "_featurestore", ""
+        )
+
+        return training_dataset_metadata
```
bubriks, line 216: It isn't a good practice to do it like this. The dictionary can easily have misspelled ids, class might be better. Furthermore, from what I see here, you only need something like a dictionary with the following keys: `path` and `query`. By just having these 2 keys it would be simplified and there would no longer be duplications between hsfs and flyingduck (for composing `path`).

python/hsfs/core/arrow_flight_client.py
```diff
@@ -0,0 +1,335 @@
...
+        filter_expression = {}
+        filter_expression["type"] = "logic"
+        filter_expression["logic_type"] = logic._type
+        if logic._left_f:
+            left_filter = self._resolve_filter(logic._left_f, featuregroups)
+        elif logic._left_l:
+            left_filter = self._resolve_logic(logic._left_l, featuregroups)
+        else:
+            left_filter = None
+        filter_expression["left_filter"] = left_filter
```
bubriks, line 307: duplication of logic with this: https://github.com/logicalclocks/feature-store-api/pull/973/files#diff-a081d1539ddd7fe9470d2077edfc0c45f48fd9db17d078799ac14c952c7e0df7R309-R315 maybe make a method and pass the different params as arguments

bubriks (changes requested):

python/hsfs/core/arrow_flight_client.py
tdoehmen: I know it would be nice, but `read_query` needs to call `_get_query_object` as well, and it will get the query and the query string as separate objects.

tdoehmen: I will try to do some more refactoring.

python/hsfs/core/arrow_flight_client.py
tdoehmen: This is an object that I will send to the backend, having it as a dict makes it easier to send and decode on the receiver side.

bubriks: Instead of sending this object can't we just send path and query?

tdoehmen: Yes you are right, it might even make sense to not construct it manually but to use `trainingdataset.location` instead. However, I'd prefer to change it when we actually hook this functionality up to hsfs. Currently it is not used, so I'd rather simply remove it for now.

python/tests/engine/test_python.py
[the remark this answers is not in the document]
tdoehmen: Tests added

bubriks (approved):

tdoehmen:
Only supports internal clients for now. @moritzmeister Further comments can be as part of the upcoming PR for external client support.

PR #981, opened 2023-04-06
[FSTORE-751] Flink support for HSFS java client

davitbzh:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/java/flink/src/main/java/com/logicalclocks/hsfs/flink/HopsworksConnection.java
+++ b/java/flink/src/main/java/com/logicalclocks/hsfs/flink/HopsworksConnection.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2022. Hopsworks AB
+ *  Copyright (c) 2023. Hopsworks AB
  *
  *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
--- a/java/flink/src/main/java/com/logicalclocks/hsfs/flink/HopsworksConnection.java
+++ b/java/flink/src/main/java/com/logicalclocks/hsfs/flink/HopsworksConnection.java
@@ -1,0 +1,78 @@
+/*
+ *  Copyright (c) 2022. Hopsworks AB
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ *  See the License for the specific language governing permissions and limitations under the License.
+ *
+ */
+
+package com.logicalclocks.hsfs.flink;
+
+import com.logicalclocks.hsfs.FeatureStoreException;
+import com.logicalclocks.hsfs.HopsworksConnectionBase;
+import com.logicalclocks.hsfs.SecretStore;
+import com.logicalclocks.hsfs.metadata.HopsworksClient;
+
+import lombok.Builder;
+
+import software.amazon.awssdk.regions.Region;
+
+import java.io.IOException;
+
+public class HopsworksConnection extends HopsworksConnectionBase {
+
+  @Builder
+  public HopsworksConnection(String host, int port, String project, Region region, SecretStore secretStore,
+                             boolean hostnameVerification, String trustStorePath,
+                             String certPath, String apiKeyFilePath, String apiKeyValue)
+      throws IOException, FeatureStoreException {
+    this.host = host;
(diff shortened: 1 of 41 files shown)


java/flink/src/main/java/com/logicalclocks/hsfs/flink/HopsworksConnection.java
```diff
@@ -0,0 +1,78 @@
...
+
+import lombok.Builder;
+
+import software.amazon.awssdk.regions.Region;
+
+import java.io.IOException;
+
+public class HopsworksConnection extends HopsworksConnectionBase {
+
+  @Builder
```
SirOibaf, line 33: For whatever reason in my example I can't use the builder pattern. I have to create the full HopsworksConnection manually. 

Looks like the builder is not public so I can't use it from outside.

java/flink/src/main/java/com/logicalclocks/hsfs/flink/engine/FlinkEngine.java
```diff
@@ -0,0 +1,94 @@
...
+
+    DataStream<Object> genericDataStream = (DataStream<Object>) dataStream;
+    Properties properties = getKafkaProperties(streamFeatureGroup);
+
+    KafkaSink<Object> sink = KafkaSink.<Object>builder()
+        .setBootstrapServers(properties.getProperty("bootstrap.servers"))
+        .setKafkaProducerConfig(properties)
+        .setRecordSerializer(KafkaRecordSerializationSchema.builder()
+        .setTopic(streamFeatureGroup.getOnlineTopicName())
+        .setValueSerializationSchema(new PojoSerializer())
```
SirOibaf, line 71: We need to properly serialize the key here as well. We need to distribute the messages on the different partitions based on the primary keys.  Here we should be setting `.setKeySerializationSchema()` I believe.

java/flink/src/main/java/com/logicalclocks/hsfs/flink/engine/PojoSerializer.java
```diff
@@ -0,0 +1,59 @@
...
+
+  @SneakyThrows
+  @Override
+  public byte[] serialize(Object input) {
+    return encode(input);
+  }
+
+  private byte[] encode(Object input) throws IOException {
+
+    Schema schema = ReflectData.get().getSchema(input.getClass());
```
SirOibaf, line 43: We discussed on Slack. This should be using the feature group avro schema, so that messages can be validated before being sent into Kafka.

Additionally this code does not account for complex features (i.e. arrays) which we have to serialize twice.

SirOibaf (changes requested):
Beside the comments I left, my general take is that if a method is not supported  and does not make sense to support in the future we should remove it.

Currently the FeatureStore.java object in the Flink package has tons of `not supported exception`. These exceptions are not documented though. So users will try to use the methods. 

Imho we should remove those methods and keep them only on the Spark package.

Same applies to the other classes like StreamFeatureGroup, FeatureView and so on.

java/flink/src/main/java/com/logicalclocks/hsfs/flink/HopsworksConnection.java
davitbzh: hmm, not sure what is the problem. could not reproduce. I could call this outside

java/flink/src/main/java/com/logicalclocks/hsfs/flink/engine/FlinkEngine.java
davitbzh: write KeySerializationSchema, we can discuss if there is more efficient way to do that. I am mostly thinking what happens when there is multiple primary keys

java/flink/src/main/java/com/logicalclocks/hsfs/flink/engine/PojoSerializer.java
davitbzh: done

SirOibaf (approved):

java/flink/src/main/java/com/logicalclocks/hsfs/flink/engine/PojoToAvroRecord.java
```diff
@@ -0,0 +1,139 @@
...
+  @Override
+  public TypeInformation<GenericRecord> getProducedType() {
+    return producedType;
+  }
+
+  private void populateAvroRecord(GenericRecord record, String fieldName, Object fieldValue) throws IOException {
+    if (this.deserializedComplexFeatureSchemas.containsKey(fieldName)) {
+      GenericDatumWriter<Object> complexFeatureDatumWriter =
+          new GenericDatumWriter<>(this.deserializedComplexFeatureSchemas.get(fieldName));
+      ByteArrayOutputStream complexFeatureByteArrayOutputStream = new ByteArrayOutputStream();
```
SirOibaf, line 122: I think we are leaking resources here - i think needs to be closed

PR #1008, opened 2023-05-09
[FSTORE-836] Show "Reading data with..." message and remove Hive warning

tdoehmen:
This PR:
- adds an animated printout "Reading data with FlyingDuck..."/ "Reading data with Hive..." in python-only client
- prints "Finished ... (0.2s)" with runtime when finished
- removed printed query and warnings when querying with hive.

JIRA Issue: FSTORE-836

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/engine/python.py
+++ b/python/hsfs/engine/python.py
@@ -21,6 +21,7 @@ import time
 import re
 import ast
 import warnings
+import logging
 import avro
 import socket
 import pyarrow as pa
@@ -65,6 +66,9 @@ from hsfs.client import exceptions, hopsworks
 from hsfs.feature_group import FeatureGroup
 from thrift.transport.TTransport import TTransportException
 from pyhive.exc import OperationalError
+
+# Disable pyhive INFO logging
+logging.getLogger("pyhive").setLevel(logging.WARNING)
 
 HAS_FAST = False
 try:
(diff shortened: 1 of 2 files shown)


python/hsfs/engine/python.py
```diff
@@ -121,12 +125,24 @@ def _sql_offline(
...
+                arrow_flight_client.get_instance().read_query,
+                sql_query,
+            )
         else:
             with self._create_hive_connection(
                 feature_store, hive_config=hive_config
             ) as hive_conn:
-                result_df = pd.read_sql(sql_query, hive_conn)
+                # Suppress SQLAlchemy pandas warning
+                with warnings.catch_warnings():
```
moritzmeister, line 138: Can we please not do this? I usually keep an eye on warnings for things that will be deprecated soon.

python/hsfs/util.py
```diff
@@ -300,6 +303,38 @@ def translate_legacy_spark_type(output_type):
...
 
+def _loading_animation(message, stop_event):
+    for char in itertools.cycle([".", "..", "...", ""]):
+        if stop_event.is_set():
+            break
+        print(f"{message}{char}   ", end="\r")
+        time.sleep(0.5)
+
+
+def run_with_loading_animation(message, func, *args, **kwargs):
```
moritzmeister, line 314: Can't we just use TQDM?

python/hsfs/engine/python.py
tdoehmen: It's about this warning in particular: "UserWarning: pandas only support SQLAlchemy connectable(engine/connection) ordatabase string URI or sqlite3 DBAPI2 connectionother DBAPI2 objects are not tested, please consider using SQLAlchemy", we should get rid of it.

Deprecation warning ragarding the connection will still show up, like this one here, that's why I didn't include it in the block.
image

tdoehmen: we can specific the filter to only filter user warning `warnings.simplefilter("ignore", UserWarning),` then deprecation warning will still show. I'll make a commit.

PR #1011, opened 2023-05-12
[FSTORE-854] Accept datetime object in filter and support unbinded features

tdoehmen:
This PR adds:
- support for datetime objects in flyingduck query filters
- support for unbinded features (i.e. `Feature("name")`)

JIRA Issue: FSTORE-854, FSTORE-853

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [x] Unit Tests
- [ ] Integration Tests
- [x] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/core/arrow_flight_client.py
+++ b/python/hsfs/core/arrow_flight_client.py
@@ -14,6 +14,7 @@
 #   limitations under the License.
 #
 
+import datetime
 import json
 import base64
 import warnings
@@ -305,10 +306,14 @@ class ArrowFlightClient:
             return None
 
     def _resolve_filter(self, filter, featuregroups):
+        if isinstance(filter._value, datetime.datetime):
+            filter_value = filter._value.strftime("%Y-%m-%d %H:%M:%S")
+        else:
+            filter_value = filter._value
         return {
             "type": "filter",
             "condition": filter._condition,
-            "value": filter._value,
+            "value": filter_value,
             "feature": f"{featuregroups[filter._feature._feature_group_id]}.{filter._feature._name}",
             "numeric": (
                 filter._feature._type in ArrowFlightClient.FILTER_NUMERIC_TYPES
--- a/python/hsfs/core/arrow_flight_client.py
+++ b/python/hsfs/core/arrow_flight_client.py
@@ -309,8 +309,8 @@ class ArrowFlightClient:
         featuregroup_name = None
 
         if feature._feature_group_id is None:
-            for fg_name in features:
-                if feature._name in features[fg_name]:
+            for fg_name, fg_features in features.items():
+                if feature._name in fg_features:
                     featuregroup_name = fg_name
                     break
         elif feature._feature_group_id in featuregroups:
--- a/python/hsfs/core/arrow_flight_client.py
+++ b/python/hsfs/core/arrow_flight_client.py
@@ -236,7 +236,7 @@ class ArrowFlightClient:
             features,
             filters,
         ) = self._collect_featuregroups_features_and_filters_rec(query)
-        filters = self._filter_to_expression(filters, featuregroups)
+        filters = self._filter_to_expression(filters, featuregroups, features)
         for feature in features:
             features[feature] = list(features[feature])
         return featuregroups, features, filters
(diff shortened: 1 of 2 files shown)


python/tests/core/test_arrow_flight_client.py
```diff
@@ -340,3 +342,82 @@ def test_construct_query_object(self, mocker, backend_fixtures):
...
+                    "condition": "GREATER_THAN",
+                    "value": "2011-03-01 12:57:02",
+                    "feature": "test.fg_test_1.intt",
+                    "numeric": False,
+                },
+                "right_filter": None,
+            },
+        }
+
+        diff = self._find_diff(query_object, query_object_reference)
```
bubriks, line 382: why do we need `_find_diff` method it might not work and cause the test to fail/succeed when it shouldn't (basically forcing us to test that method as well)? cant we just use `==` or something like `DeepDiff`?

python/hsfs/core/arrow_flight_client.py
```diff
@@ -279,40 +280,61 @@ def _collect_featuregroups_features_and_filters_rec(self, query):
 
         return featuregroups, features, filters
 
-    def _filter_to_expression(self, filters, featuregroups):
+    def _filter_to_expression(self, filters, featuregroups, features):
```
bubriks, line 283: i don't doubt it works, i just think it would make sense to have features be part of the feature group (instead of a separate dictionary). Now the FG name is the link between the two dictionaries.

Also why not use existing FG and Feature classes?

python/tests/core/test_arrow_flight_client.py
tdoehmen: yep think I could simplify it

python/hsfs/core/arrow_flight_client.py
tdoehmen: The problem this is PR is trying to solve is that some features don't have an associated featuregroup, I am establishing the association here.

tdoehmen: I agree though that it could be done prettier but in the end I need them as separate lists/dicts in exactly that format to send them to the fd backend.

bubriks: yeah, right now it looks like this
```
"featuregroups": {15: "test.fg_test_1"}
"features": {"test.fg_test_1": ["intt", "stringt"]}
```
but it might be better like this:
```
"featuregroups": [{"id": 15, "name": "test.fg_test_1", "features": ["intt", "stringt"]}]
```

bubriks (commented):
some more minor comments

python/hsfs/core/arrow_flight_client.py
```diff
@@ -279,40 +280,61 @@ def _collect_featuregroups_features_and_filters_rec(self, query):
...
+        if featuregroup_name is None:
+            raise FeatureStoreException(f"Feature {feature._name} not found in query")
+
+        return f"{featuregroup_name}.{feature._name}"
+
+    def _resolve_filter(self, filter, featuregroups, features):
+        if isinstance(filter._value, datetime.datetime):
+            filter_value = filter._value.strftime("%Y-%m-%d %H:%M:%S")
+        else:
+            filter_value = filter._value
```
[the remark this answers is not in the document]
tdoehmen: Not sure, in that case it's assgning the value twice when it goes into the if.

python/hsfs/core/arrow_flight_client.py
```diff
@@ -279,40 +280,61 @@ def _collect_featuregroups_features_and_filters_rec(self, query):
...
         else:
             return None
 
-    def _resolve_filter(self, filter, featuregroups):
+    def _get_full_feature_name(self, feature, featuregroups, features):
+        featuregroup_name = None
+
+        if feature._feature_group_id is None:
+            for fg_name in features:
+                if feature._name in features[fg_name]:
```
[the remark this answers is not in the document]
tdoehmen: fine by me

bubriks (approved):

python/hsfs/core/arrow_flight_client.py
tdoehmen: agree, but it's a bit out of scope for this PR

PR #1019, opened 2023-05-17
[FSTORE-848] Adding more unit tests

vatj:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


python/hsfs/core/monitoring_window_config_engine.py
```diff
@@ -243,3 +249,193 @@ def get_window_start_end_times(
...
+        Returns:
+            `pyspark.sql.DataFrame`. A Spark DataFrame with the entity data
+        """
+        if use_event_time:
+            entity_df = entity._feature_view_engine.get_batch_query(
+                feature_view_obj=entity,
+                start_time=start_time,
+                end_time=end_time,
+                with_label=True if feature_name is None else False,
+                training_dataset_version=transformation_function_dataset_version,
```
javierdlrm, line 397: Have you checked what happens if training_dataset_version = None ?

python/hsfs/core/monitoring_window_config_engine.py
```diff
@@ -243,3 +249,193 @@ def get_window_start_end_times(
...
+
+        Returns:
+            `pyspark.sql.DataFrame`. A Spark DataFrame with the entity data
+        """
+        if use_event_time:
+            entity_df = entity._feature_view_engine.get_batch_query(
+                feature_view_obj=entity,
+                start_time=start_time,
+                end_time=end_time,
+                with_label=True if feature_name is None else False,
```
javierdlrm, line 396: If the feature_name is the label, it won't be present in the entity_df. Why not setting always to True?

python/hsfs/core/feature_monitoring_config.py
```diff
@@ -59,20 +62,24 @@ def __init__(
         feature_view_name: Optional[str] = None,
         feature_view_version: Optional[int] = None,
         href: Optional[str] = None,
-    ):
+        transformation_function_dataset_version: Optional[int] = None,
```
javierdlrm, line 65: I think it's better to stick to `training_dataset_version` instead of renaming it. since we don't have such concept of transformation function dataset.

Alternatively, maybe `transformation_functions_td_dataset_version`, although it's a bit verbose. I think `training_dataset_version`  is fine.

python/hsfs/core/monitoring_window_config_engine.py
vatj: Fair enough

python/hsfs/core/monitoring_window_config_engine.py
vatj: Ja, this version is used to add extra filters that could have been added when creating the training dataset. So if you pass none it just adds no extra filter

PR #1040, opened 2023-06-09
[FSTORE-893] Get Hopsworks certificates for external Flink client 

davitbzh:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/java/hsfs/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksExternalClient.java
+++ b/java/hsfs/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksExternalClient.java
@@ -96,21 +96,21 @@ public class HopsworksExternalClient implements HopsworksHttpClient {
                                  boolean hostnameVerification, String trustStorePath)
       throws IOException, FeatureStoreException, KeyStoreException, CertificateException,
       NoSuchAlgorithmException, KeyManagementException {
-    this(host, port, null, null, hostnameVerification, trustStorePath, apiKeyFilepath, null);
+    this(host, port, null, null, hostnameVerification, trustStorePath, apiKeyFilepath, null, null);
   }
 
   public HopsworksExternalClient(String host, int port, boolean hostnameVerification,
                                  String trustStorePath, Region region, SecretStore secretStore)
       throws IOException, FeatureStoreException, KeyStoreException, CertificateException,
       NoSuchAlgorithmException, KeyManagementException {
-    this(host, port, region, secretStore, hostnameVerification, trustStorePath, null, null);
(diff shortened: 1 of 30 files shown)


java/hsfs/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksExternalClient.java
```diff
@@ -96,21 +96,21 @@ public HopsworksExternalClient(String host, int port, String apiKeyFilepath,
                                  boolean hostnameVerification, String trustStorePath)
       throws IOException, FeatureStoreException, KeyStoreException, CertificateException,
       NoSuchAlgorithmException, KeyManagementException {
-    this(host, port, null, null, hostnameVerification, trustStorePath, apiKeyFilepath, null);
+    this(host, port, null, null, hostnameVerification, trustStorePath, apiKeyFilepath, null, null);
```
SirOibaf, line 99: I don't see these methods being used anywhere. Should we remove them?

java/hsfs/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksExternalClient.java
```diff
@@ -119,11 +119,12 @@ public HopsworksExternalClient(CloseableHttpClient httpClient, HttpHost httpHost
   }
 
   HopsworksExternalClient(String host, int port, Region region,
-                          SecretStore secretStore, boolean hostnameVerification,
+                          SecretStore secretStore, boolean hostnameVerification, String keyStorePath,
```
SirOibaf, line 122: It's not clear to me why we need the keystore here. At best, the http client only needs the truststore (only if Hopsworks is using a self signed certificate). It doesn't need the keystore.

Also in the `flink/HopsworksConnection.java` we are overwriting this path anyway with the certificate path from the Flink configuration. Same happens in the Spark client.

I really don't think we need to pass the keyStorePath here.

SirOibaf (changes requested):

java/hsfs/src/main/java/com/logicalclocks/hsfs/metadata/HopsworksExternalClient.java
davitbzh: agreed. I was thinking if we should ever let users to pass certificate path from here. I will remove it now

SirOibaf (approved):

PR #1064, opened 2023-06-27
[FSTORE-908] Upgrade the job scheduler to use cron expression

vatj:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/core/feature_monitoring_config.py
+++ b/python/hsfs/core/feature_monitoring_config.py
@@ -477,6 +477,9 @@ class FeatureMonitoringConfig:
             my_monitoring_config.run_job()
             ```
 
+        !!! info
+            The job will be triggered asynchronously and the method will return immediately.
+
         # Raises
             `FeatureStoreException`: If the feature monitoring config has not been saved.
 
--- a/python/hsfs/core/feature_monitoring_config.py
+++ b/python/hsfs/core/feature_monitoring_config.py
@@ -24,9 +24,12 @@ from hsfs.client.exceptions import FeatureStoreException
 
 from hsfs.core import monitoring_window_config as mwc
 from hsfs.core.feature_monitoring_result import FeatureMonitoringResult
-from hsfs.core import monitoring_window_config_engine
-from hsfs.core import feature_monitoring_config_engine
-from hsfs.core import feature_monitoring_result_engine
+from hsfs.core import (
+    monitoring_window_config_engine,
+    feature_monitoring_config_engine,
+    feature_monitoring_result_engine,
+    job_scheduler_engine,
+)
 
 MAX_LENGTH_NAME = 63
 MAX_LENGTH_DESCRIPTION = 2000
@@ -66,7 +69,7 @@ class FeatureMonitoringConfig:
         href: Optional[str] = None,
         transformed_with_version: Optional[int] = None,
         is_event_time: bool = False,
-    ) -> "FeatureMonitoringConfig":
+    ):
         self.name = name
         self._id = id
         self._href = href
@@ -81,6 +84,7 @@ class FeatureMonitoringConfig:
         self._transformed_with_version = transformed_with_version
         self._is_event_time = is_event_time
 
+        self._job_scheduler_engine = job_scheduler_engine.JobSchedulerEngine()
         self._feature_monitoring_config_engine = (
             feature_monitoring_config_engine.FeatureMonitoringConfigEngine(
                 feature_store_id=feature_store_id,
@@ -313,7 +317,7 @@ class FeatureMonitoringConfig:
             `FeatureMonitoringConfig`. The updated FeatureMonitoringConfig object.
         """
         self.reference_window_config = {
-            "specific_value": float(value),
+            "specific_value": value,
         }
 
         return self
@@ -479,7 +483,7 @@ class FeatureMonitoringConfig:
         # Returns
             `Job`. A handle for the job computing the statistics.
         """
-        if not self._id:
+        if not self._id or not self._job_name:
             raise FeatureStoreException(
                 "Feature monitoring config must be registered via `.save()` before computing statistics."
             )
@@ -509,7 +513,7 @@ class FeatureMonitoringConfig:
         # Returns
             `Job`. A handle for the job computing the statistics.
         """
-        if not self._id:
+        if not self._id or not self._job_name:
             raise FeatureStoreException(
                 "Feature monitoring config must be registered via `.save()` before fetching"
                 "the associated job."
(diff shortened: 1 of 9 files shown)


python/hsfs/core/feature_monitoring_config.py
```diff
@@ -479,7 +483,7 @@ def run_job(self):
         # Returns
             `Job`. A handle for the job computing the statistics.
         """
-        if not self._id:
+        if not self._id or not self._job_name:
```
javierdlrm, line 486: Could it be that the FM config is created but doesn't have a job attached to it?
If so, we should update the exception message accordingly.

python/hsfs/core/job_scheduler.py
```diff
@@ -14,25 +14,34 @@
...
 
 
 class JobScheduler:
     def __init__(
         self,
-        start_date_time: Union[int, str, datetime, pd.Timestamp],
-        job_frequency: str,
+        start_date_time: Union[int, str, datetime, date, pd.Timestamp],
+        job_frequency: Optional[str] = "CUSTOM",
+        cron_expression: Optional[str] = None,
```
javierdlrm, line 32: To keep it simpler, have you thought of merging these two parameters?
Something like job_frequency with possible values "HOURLY", "DAILY", "WEEKLY" or "/cron expression/".

python/hsfs/core/job_scheduler.py
```diff
@@ -45,45 +54,255 @@ def __init__(
...
+        return self._job_scheduler_engine.update_job_scheduler(the_job_scheduler=self)
+
+    def pause(self):
+        """Pauses the scheduling of job in Hopsworks, the job can still be triggered manually."""
+        if self._id:
+            raise ValueError(
+                "Cannot pause a scheduler not registered, use `save()` to register a new scheduler to Hopsworks."
+            )
+        if not self._job_name:
+            raise ValueError("Cannot pause a scheduler without a job name.")
```
javierdlrm, line 131: Could this be actually possible? An `existing` job scheduler without a job name?

python/hsfs/core/job_scheduler.py
```diff
@@ -45,45 +54,255 @@ def __init__(
...
+
+        self._job_scheduler_engine.pause_or_resume_job_scheduler(
```
javierdlrm, line 133: I may be a bit picky here, but could we have separate methods `pause_job_scheduler` and `resume_job_scheduler` instead of the same method with a non-optional boolean `pause` parameter?

python/hsfs/feature_group.py
```diff
@@ -1343,6 +1354,11 @@ def _create_feature_monitoring(
                 Options are "HOURLY", "DAILY", "WEEKLY", "MONTHLY", defaults to "DAILY".
```
javierdlrm, line 1354: In the JobScheduler constructor, it seems MONTHLY it is not supported, only HOURLY, DAILY and WEEKLY.

python/hsfs/feature_view.py
```diff
@@ -2476,6 +2478,11 @@ def _create_statistics_monitoring(
                 Options are "HOURLY", "DAILY", "WEEKLY", "MONTHLY", defaults to "DAILY".
```
javierdlrm, line 2478: Same as previous comment.
In the JobScheduler constructor, it seems MONTHLY it is not supported, only HOURLY, DAILY and WEEKLY.

python/hsfs/feature_view.py
```diff
@@ -2553,6 +2564,11 @@ def _create_feature_monitoring(
...
```
javierdlrm, line 2564: Same as previous comment.
In the JobScheduler constructor, it seems MONTHLY it is not supported, only HOURLY, DAILY and WEEKLY.

python/hsfs/feature_group.py
```diff
@@ -1270,9 +1272,14 @@ def _create_statistics_monitoring(
             feature_name: Name of the feature to monitor. If not specified, statistics
                 will be computed for all features.
             job_frequency: Frequency at which to compute the statistics for the feature.
-                Options are "HOURLY", "DAILY", "WEEKLY", "MONTHLY", defaults to "DAILY".
+                Options are "NEAR REAL-TIME", "HOURLY", "DAILY", "WEEKLY", defaults to "DAILY".
```
javierdlrm, line 1275: What is NEAR REAL-TIME? How is it handled?
In the JobScheduler constructor, it seems only HOURLY, DAILY and WEEKLY are supported

python/hsfs/core/job_scheduler_engine.py
```diff
@@ -0,0 +1,174 @@
...
+
+        if job_frequency and job_frequency.upper() not in [
+            "HOURLY",
+            "DAILY",
+            "WEEKLY",
+            "CUSTOM",
+        ]:
+            raise ValueError(
+                "Invalid job frequency. Supported frequencies are HOURLY, DAILY, WEEKLY."
+            )
```
javierdlrm, line 80: This is already checked in the JobScheduler constructor

python/hsfs/core/feature_monitoring_config.py
vatj: In principle no, there should not be a case where it is registered and there is no job name. The message is not specific to id or job name in that regard so I think it is fine.

python/hsfs/core/feature_monitoring_config_engine.py
```diff
@@ -320,6 +273,8 @@ def _build_default_statistics_monitoring_config(
             scheduler_config={
                 "job_frequency": job_frequency,
                 "start_date_time": start_date_time,
+                "end_date_time": end_date_time,
+                "cron_expression": cron_expression,
                 "enabled": True,
             },
```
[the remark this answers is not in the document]
vatj: Scheduler config goes through the setter on init so that it gets converted to a jobSchedule instance with validation :)

python/hsfs/core/job_scheduler.py
vatj: Yes I thought about it. I think it is something you and the rest of the team will have to decide when finalising docs and tutorial. I definitely agree think it is easier UX. I can fix it now if you think it is a no-brainer

python/hsfs/core/job_scheduler.py
vatj: I am not sure but since the job name is basically used to query the rest api all the time I would say we keep a lot of checks in place to avoid issues

PR #1068, opened 2023-07-03
[FSTORE-939] Change name of Stream Feature Group's DeltaStreamer Job

davitbzh:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/engine/python.py
+++ b/python/hsfs/engine/python.py
@@ -1312,4 +1312,4 @@ class Engine:
             else:
                 return True
         else:
-            return False
+            return True
--- a/python/hsfs/engine/python.py
+++ b/python/hsfs/engine/python.py
@@ -970,10 +970,7 @@ class Engine:
             and reset_offsets
             and not feature_group._multi_part_insert
         ):
-            if offline_write_options is not None and not (
-                offline_write_options.get("start_offline_materialization", True)
-                or offline_write_options.get("start_offline_backfill", True)
-            ):
+            if self._start_offline_materialization(offline_write_options):
                 warnings.warn(
                     "This is the first ingestion after an upgrade or backup/restore, running materialization job even though `start_offline_materialization` was set to `False`.",
                     util.FeatureGroupWarning,
@@ -983,14 +980,9 @@ class Engine:
                 + " -kafkaOffsetReset true",
                 await_termination=offline_write_options.get("wait_for_job", False),
             )
-        elif (
-            not isinstance(feature_group, ExternalFeatureGroup)
-            and offline_write_options is not None
-            and (
-                offline_write_options.get("start_offline_materialization", True)
-                or offline_write_options.get("start_offline_backfill", True)
-            )
-        ):
+        elif not isinstance(
+            feature_group, ExternalFeatureGroup
+        ) and self._start_offline_materialization(offline_write_options):
             feature_group.materialization_job.run(
                 await_termination=offline_write_options.get("wait_for_job", False)
             )
@@ -1309,3 +1301,13 @@ class Engine:
             StorageConnector.SNOWFLAKE,
             StorageConnector.KAFKA,
         ]
+
+    @staticmethod
+    def _start_offline_materialization(offline_write_options):
+        if offline_write_options is not None:
+            if "start_offline_materialization" in offline_write_options:
+                return offline_write_options.get("start_offline_materialization", True)
+            elif "start_offline_backfill" in offline_write_options:
+                return offline_write_options.get("start_offline_backfill", True)
+        else:
+            return False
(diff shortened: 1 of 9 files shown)


python/hsfs/engine/python.py
```diff
@@ -963,36 +963,40 @@ def acked(err, msg):
...
         if (
             not isinstance(feature_group, ExternalFeatureGroup)
             and reset_offsets
             and not feature_group._multi_part_insert
         ):
-            if offline_write_options is not None and not offline_write_options.get(
-                "start_offline_backfill", True
+            if offline_write_options is not None and not (
+                offline_write_options.get("start_offline_materialization", True)
+                or offline_write_options.get("start_offline_backfill", True)
```
SirOibaf, line 975: This if statement doesn't work. Users will only set one of the two (Let's say `start_offline_materialization` to `False`). The other one is going to be empty and so the results is going to be True.

I think the easiest is to remove the default value and check if the keys exists.

python/hsfs/engine/python.py
```diff
@@ -963,36 +963,40 @@ def acked(err, msg):
...
             ):
                 warnings.warn(
-                    "This is the first ingestion after an upgrade or backup/restore, running backfill job even though `start_offline_backfill` was set to `False`.",
+                    "This is the first ingestion after an upgrade or backup/restore, running materialization job even though `start_offline_materialization` was set to `False`.",
                     util.FeatureGroupWarning,
                 )
-            feature_group.backfill_job.run(
-                args=feature_group.backfill_job.config.get("defaultArgs", "")
+            feature_group.materialization_job.run(
```
SirOibaf, line 981: Isn't this misaligned? Looking at the code (assuming the if statement above is fixed), if the user doesn't want to start the job, we only print a warning, but then we keep going and we actually run the job nevertheless?

Or am I missign somthing?

python/hsfs/engine/python.py
```diff
@@ -963,36 +963,40 @@ def acked(err, msg):
...
                 await_termination=offline_write_options.get("wait_for_job", False),
             )
         elif (
             not isinstance(feature_group, ExternalFeatureGroup)
             and offline_write_options is not None
-            and offline_write_options.get("start_offline_backfill", True)
+            and (
+                offline_write_options.get("start_offline_materialization", True)
+                or offline_write_options.get("start_offline_backfill", True)
+            )
```
SirOibaf, line 992: Same comemnt as above.

SirOibaf (changes requested):
Generally speaking I would mention in the documentation that the `start_offline_backfill` is deprecated but still works.

python/hsfs/engine/python.py
davitbzh: hmm, checking master and indentation is the same there. But the warning saying `running materialization job even though `start_offline_materialization` was set to `False`` I guess that was intended?

python/hsfs/feature_group.py
```diff
@@ -1663,8 +1663,14 @@ def save(
                   whether or not to the save call should return only
                   after the Hopsworks Job has finished. By default it does not wait.
                 * key `start_offline_backfill` and value `True` or `False` to configure
-                  whether or not to start the backfill job to write data to the offline
-                  storage. By default the backfill job gets started immediately.
+                  whether or not to start the materialization job to write data to the offline
+                  storage. `start_offline_backfill` is deprecated. Use `start_offline_materialization` instead.
+                * key `start_offline_backfill` and value `True` or `False` to configure
+                  whether or not to start the materialization job to write data to the offline
+                  storage. `start_offline_backfill` is deprecated. Use `start_offline_materialization` instead.
```
SirOibaf, line 1670: Duplicated documentation

python/hsfs/engine/python.py
SirOibaf: Ok I see what you mean but then the indentation is correct.

python/hsfs/engine/python.py
```diff
@@ -1305,3 +1301,15 @@ def is_connector_type_supported(connector_type):
...
+    def _start_offline_materialization(offline_write_options):
+        if offline_write_options is not None:
+            if "start_offline_materialization" in offline_write_options:
+                return offline_write_options.get("start_offline_materialization")
+            elif "start_offline_backfill" in offline_write_options:
+                return offline_write_options.get("start_offline_backfill")
+            else:
+                return True
+        else:
+            return False
```
SirOibaf, line 1315: Shouldn't this be returning True? if there are no options, the default behavior is to start the materialization job.

SirOibaf (changes requested):

SirOibaf (approved):

PR #1069, opened 2023-07-05
[FSTORE-903] Snowflake Storage Connector

tdoehmen:
This PR adds Snowflake storage connector support for flyingduck

JIRA Issue: FSTORE-903

Priority for Review: -

Related PRs: https://github.com/logicalclocks/flyingduck/pull/45 

**How Has This Been Tested?**

- [x] Unit Tests
- [ ] Integration Tests
- [x] Manual Tests on VM

```
```


python/hsfs/core/arrow_flight_client.py
```diff
@@ -203,51 +208,82 @@ def read_path(self, path):
...
+            connector["type"] = "hudi"
+
+        return connector
+
     def _serialize_featuregroup_name(self, fg):
         return f"{fg._get_project_name()}.{fg.name}_{fg.version}"
 
-    def _serialize_filter_expression(self, query):
-        if query.filters is None:
+    def _serialize_filter_expression(self, filters, query, short_name=False):
```
bubriks, line 261: why was it changed? To me it looks like previously it was better since now we always have to extract `query._filter, query` while we could just pass `query`

tdoehmen: Since we have multiple different sources now (snowflake, hudi), we need to handle pushdowns per featuregroup. If you have a join between a hudi and a snowflake table, we need to push down the filter that applies to the snowflake table (s. https://github.com/logicalclocks/feature-store-api/pull/1069/files#diff-a081d1539ddd7fe9470d2077edfc0c45f48fd9db17d078799ac14c952c7e0df7R221). I still wanted to re-use that method though.

tdoehmen: Indeed, I think we change it back, I think it was some intermediate state where i needed it like this, but looks like it can be reverted again.

python/hsfs/feature_group.py
```diff
@@ -2630,6 +2630,11 @@ def __init__(
 
         self._href = href
 
+    def _get_project_name(self):
+        if self.feature_store_name.endswith("_featurestore"):
+            return self.feature_store_name[:-13]
```
bubriks, line 2635:
```suggestion
        feature_store_suffix = "_featurestore"
        if self.feature_store_name.endswith(feature_store_suffix):
            return self.feature_store_name[:-len(feature_store_suffix)]
```

python/hsfs/feature_group.py
```diff
@@ -2743,15 +2748,17 @@ def read(
...
+        #
+        # if engine.get_type() == "python" and not online:
+        #    raise FeatureStoreException(
+        #        "Reading an External Feature Group directly into a Pandas Dataframe using "
+        #        + "Python/Pandas as Engine from the external storage system "
+        #        + "is not supported, however, if the feature group is online enabled, you can read "
+        #        + "from online storage or you can use the "
+        #        + "Query API to create Feature Views/Training Data containing External "
+        #         + "Feature Groups."
+        #    )
```
bubriks, line 2761:
```suggestion
```

python/hsfs/feature_group.py
tdoehmen: moved it to util

python/hsfs/feature_group.py
tdoehmen: I make it print the regular error message if the external fg not supported by flyingduck or if flyingduck is not enabled/reachable

python/hsfs/util.py
```diff
@@ -62,14 +64,21 @@ def feature_group_name(feature_group):
 
 
 def rewrite_feature_store_name(name):
```
bubriks, line 66: maybe rename it to `append_feature_store_suffix`

python/hsfs/util.py
```diff
@@ -62,14 +64,21 @@ def feature_group_name(feature_group):
...
-    FEATURE_STORE_NAME_SUFFIX = "_featurestore"
     name = name.lower()
     if name.endswith(FEATURE_STORE_NAME_SUFFIX):
         return name
     else:
         return name + FEATURE_STORE_NAME_SUFFIX
 
 
+def strip_feature_store_name(name):
```
bubriks, line 74: maybe rename it to `strip_feature_store_suffix`

python/hsfs/feature_group.py
```diff
@@ -2652,6 +2650,9 @@ def __init__(
 
         self._href = href
 
+    def _get_project_name(self):
```
bubriks, line 2653: maybe instead of adding new method here, we could just add it to the base fg class (then also this can be removed: https://github.com/logicalclocks/feature-store-api/pull/1069/files#diff-4ac064b1d72d926de7a5387b7d15fa9b889ac5a05ba8bcbc9974d9de334a5435R2436)

python/hsfs/core/arrow_flight_client.py
```diff
@@ -203,51 +208,82 @@ def read_path(self, path):
...
+                fg, query, on_demand_fg_aliases
+            )
             features[fg_name] = [feat.name for feat in fg.features]
-        filters = self._serialize_filter_expression(query)
-        for feature in features:
-            features[feature] = list(features[feature])
+            connectors[fg_name] = fg_connector
+        filters = self._serialize_filter_expression(query.filters, query)
+        for fg_name in features:
+            features[fg_name] = list(features[fg_name])
```
bubriks, line 223: why is this needed? Isn't `features[fg_name]` a list already?

python/tests/core/test_arrow_flight_client.py
```diff
@@ -430,6 +434,92 @@ def test_construct_query_object_without_fs_excluded(self, mocker, backend_fixtur
...
+            return_value="test",
+        )
+        query = test_fg1.filter(Feature("c_acctbal") > 500)
+
+        # Act
+        query_object = arrow_flight_client.get_instance().create_query_object(
+            query, "SELECT * FROM..."
+        )
+
+        print(query_object)
```
bubriks, line 470:
```suggestion
```

bubriks (approved):
looks good just this minor change

PR #1070, opened 2023-07-05
[FSTORE-942][FSTORE-647] adls path validation and default path fixes for path connectors

DhananjayMukhedkar:
This PR adds/fixes/changes...

JIRA Issue: -
https://hopsworks.atlassian.net/browse/FSTORE-647
https://hopsworks.atlassian.net/browse/FSTORE-942

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [x] Unit Tests
- [ ] Integration Tests
- [x] Manual Tests on VM

```
```

### DIFF
--- a/java/spark/src/main/java/com/logicalclocks/hsfs/spark/util/StorageConnectorUtils.java
+++ b/java/spark/src/main/java/com/logicalclocks/hsfs/spark/util/StorageConnectorUtils.java
@@ -75,7 +75,7 @@ public class StorageConnectorUtils {
     }
     if (path != null && !path.startsWith("s3://")) {
       path = connector.getPath(path);
-      logger.info(String.format( "Prepending default bucket specified on connector, final path: %s", path));
+      logger.info(String.format("Prepending default bucket specified on connector, final path: %s", path));
     }
     return SparkEngine.getInstance().read(connector, dataFormat, readOptions, path);
   }
@@ -111,7 +111,7 @@ public class StorageConnectorUtils {
    * @throws IOException Generic IO exception.
    */
   public Dataset<Row> read(StorageConnector.AdlsConnector connector, String dataFormat, Map<String, String> options,
-    String path) throws FeatureStoreException, IOException {
+      String path) throws FeatureStoreException, IOException {
     return SparkEngine.getInstance().read(connector, dataFormat, options, validateAdlsPath(connector, path));
   }
 
--- a/java/spark/src/main/java/com/logicalclocks/hsfs/spark/util/StorageConnectorUtils.java
+++ b/java/spark/src/main/java/com/logicalclocks/hsfs/spark/util/StorageConnectorUtils.java
@@ -18,9 +18,9 @@
 package com.logicalclocks.hsfs.spark.util;
 
 import com.google.common.base.Strings;
-import com.logicalclocks.hsfs.spark.engine.SparkEngine;
 import com.logicalclocks.hsfs.FeatureStoreException;
 import com.logicalclocks.hsfs.StorageConnector;
+import com.logicalclocks.hsfs.spark.engine.SparkEngine;
 import com.logicalclocks.hsfs.util.Constants;
 import org.apache.spark.sql.Dataset;
 import org.apache.spark.sql.Row;
@@ -32,8 +32,10 @@ import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.Base64;
 import java.util.Map;
+import java.util.logging.Logger;
 
 public class StorageConnectorUtils {
+  Logger logger = Logger.getLogger(StorageConnectorUtils.class.getName());
 
   /**
    * Reads path into a spark dataframe using the HopsFsConnector.
@@ -71,6 +73,10 @@ public class StorageConnectorUtils {
     if (options != null && !options.isEmpty()) {
       readOptions.putAll(options);
     }
+    if (path != null && !path.startsWith("s3://")) {
+      path = connector.getPath(path);
+      logger.info(String.format( "Prepending default bucket specified on connector, final path: %s", path));
+    }
     return SparkEngine.getInstance().read(connector, dataFormat, readOptions, path);
   }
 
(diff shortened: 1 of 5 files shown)


python/hsfs/storage_connector.py
```diff
@@ -587,6 +588,51 @@ def prepare_spark(self, path: Optional[str] = None):
...
+        )
+
+    def _validate_path(self, read_path: str) -> str:
+        ERROR_GEN2_PATH = "Not a valid ADLS path. For Gen2 connectors, path should follow format as 'abfss://[container-name]@[account_name].dfs.core.windows.net/[path]' "
+        ERROR_GEN1_PATH = "Not a valid ADLS path. For Gen1 connectors, path should follow format as 'adl://[account_name].azuredatalakestore.net' "
+        read_path = read_path.strip()
+        if not read_path:
+            print(
+                "No path provided. Using default path from connector, final path to read {}".format(
+                    self._get_path(read_path)
```
bubriks, line 621: instead of calling `self._get_path(read_path)` 2 times we could just save it in local variable.

python/hsfs/storage_connector.py
```diff
@@ -587,6 +588,51 @@ def prepare_spark(self, path: Optional[str] = None):
...
+        # Returns
+            `DataFrame`.
+        """
+        return engine.get_instance().read(
+            self, data_format, options, self._validate_path(path)
+        )
+
+    def _validate_path(self, read_path: str) -> str:
+        ERROR_GEN2_PATH = "Not a valid ADLS path. For Gen2 connectors, path should follow format as 'abfss://[container-name]@[account_name].dfs.core.windows.net/[path]' "
+        ERROR_GEN1_PATH = "Not a valid ADLS path. For Gen1 connectors, path should follow format as 'adl://[account_name].azuredatalakestore.net' "
```
bubriks, line 616: these messages are quite similar, maybe we could save it as a template and just insert the parts that change (ex. `Gen2` and `'abfss://[container-name]@[account_name].dfs.core.windows.net/[path]'`)

java/spark/src/main/java/com/logicalclocks/hsfs/spark/util/StorageConnectorUtils.java
```diff
@@ -278,4 +288,24 @@ public Dataset<Row> readStream(StorageConnector.KafkaConnector connector, String
...
         messageFormat.toLowerCase(), schema, options, includeMetadata);
   }
+
+  public String validateAdlsPath(StorageConnector.AdlsConnector connector, String path) throws FeatureStoreException {
+    final String ErrorGen2 =
+        "Not a valid ADLS path. For Gen2 connectors, path should follow format as "
+          + "'abfss://[container-name]@[account_name].dfs.core.windows.net/[path]' ";
+    final String ErrorGen1 =
+        "Not a valid ADLS path. For Gen1 connectors, path should follow format as "
+          + "'adl://[account_name].azuredatalakestore.net' ";
```
bubriks, line 298: could be made into a template (like in python)

java/spark/src/main/java/com/logicalclocks/hsfs/spark/util/StorageConnectorUtils.java
```diff
@@ -278,4 +288,24 @@ public Dataset<Row> readStream(StorageConnector.KafkaConnector connector, String
     return SparkEngine.getInstance().readStream(connector, connector.sparkFormat,
         messageFormat.toLowerCase(), schema, options, includeMetadata);
   }
+
+  public String validateAdlsPath(StorageConnector.AdlsConnector connector, String path) throws FeatureStoreException {
```
bubriks, line 292: why in Python this same method is part of the `AdlsConnector`, but in Java its in `StorageConnectorUtils`?

bubriks (commented):
some minor feedback, but overall looks good

python/hsfs/storage_connector.py
```diff
@@ -587,6 +588,51 @@ def prepare_spark(self, path: Optional[str] = None):
     def _get_path(self, sub_path: str):
         return os.path.join(self.path, sub_path)
 
+    def read(
+        self,
+        query: str = None,
+        data_format: str = None,
+        options: dict = {},
+        path: str = "",
+    ):
```
SirOibaf, line 597: This method is wrong. It should behave like the read method from the S3 connector https://github.com/logicalclocks/feature-store-api/blob/30443492fcabdc21a8c0f9d458c5dd7f9741bc5c/python/hsfs/storage_connector.py#L266

The `path` parameter should be the path within the ADLS container and not the full path including the container name, account id and the domain.
Otherwise there is absolutely no point of creating the storage connector if users need to provide again this information while calling the `read()` method.

Additionally the difference between Gen1 and Gen2 are not well known, so you are asking the user to know this. Again what's the point of the storage connector then?

Please take a look at the S3 implementation and do the same.

python/hsfs/storage_connector.py
SirOibaf, line 616: See my comment above. there is no point in doing validation here. Users should not be providing the full url.

SirOibaf (changes requested):

python/hsfs/storage_connector.py
DhananjayMukhedkar: Alright, looking at the docs and implementation I thought path should follow gen1 and gen2 as specified on the container option. I made changes to read from the default container specified if the path is not a full URI and removed validation for gen1 and gen2 path format.

python/hsfs/storage_connector.py
DhananjayMukhedkar: removed validation for gen1 for gen2 format.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/util/StorageConnectorUtils.java
DhananjayMukhedkar: @bubriks We had it same for java in the past. But there were some refactoring to  the return type of read method which caused to return Object instead so it was moved in another class. Davit knows better he worked on it.

python/hsfs/storage_connector.py
```diff
@@ -587,6 +587,36 @@ def prepare_spark(self, path: Optional[str] = None):
...
+            If no path is specified default container path will be used from connector.
+
+        # Returns
+            `DataFrame`.
+        """
+        path = path.strip()
+        if not path.startswith("abfss://") or path.startswith("adl://"):
+            path = self._get_path(path)
+            print(
+                "Prepending default bucket specified on connector, final path: {}".format(
```
SirOibaf, line 613: `bucket` is not a term for adls containers.

java/spark/src/main/java/com/logicalclocks/hsfs/spark/util/StorageConnectorUtils.java
SirOibaf, line 292: This needs to be fixed to match the Python behavior.

SirOibaf (changes requested):

python/hsfs/storage_connector.py
DhananjayMukhedkar: done

SirOibaf (approved):

PR #1086, opened 2023-07-25
[FSTORE-877] Materializing Training Dataset with Arrow Flight (light)

tdoehmen:
This PR adds support for materializing training datasets with arrow flight. This iteration only supports meterializing training datasets without splits and without transformation functions. When arrow flight server is installed, it will automatically be used. To explicitly use spark instead, users can set a new flag `write_options={"use_spark": True}`.

JIRA Issue: FSTORE-877

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Manual Tests on VM

```
```


python/hsfs/engine/python.py
```diff
@@ -670,6 +670,24 @@ def write_training_dataset(
                 "Currently only query based training datasets are supported by the Python engine"
             )
 
+        if (
+            arrow_flight_client.get_instance().is_query_supported(
+                dataset, user_write_options
+            )
+            and len(training_dataset.splits) == 0
+            and len(training_dataset.transformation_functions) == 0
```
kennethmhc, line 678: does flying duck support output format other than parquet?

tdoehmen: Not yet, I should filter that as well

PR #1110, opened 2023-09-07
[FSTORE-1008] enable interacting with java client to hopsworks

davitbzh:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


java/hsfs/src/main/java/com/logicalclocks/hsfs/engine/FeatureViewEngineBase.java
```diff
@@ -79,35 +79,35 @@ public static List<TrainingDatasetFeature> makeLabelFeatures(QueryBase query, Li
       for (Feature feat : (List<Feature>) query.getLeftFeatures()) {
         labelWithPrefixToFeature.put(feat.getName(), feat.getName());
         labelWithPrefixToFeatureGroup.put(feat.getName(),
-            (new FeatureGroupBaseForApi(null, feat.getFeatureGroupId())));
+            (new StreamFeatureGroup(null, feat.getFeatureGroupId())));
```
kennethmhc, line 82: why change it to `StreamFeatureGroup`?

java/hsfs/src/main/java/com/logicalclocks/hsfs/FeatureStore.java
```diff
@@ -0,0 +1,118 @@
...
+import lombok.NonNull;
+
+import java.io.IOException;
+
+public class FeatureStore extends FeatureStoreBase<Query> {
+
+  private FeatureGroupEngine featureGroupEngine;
+  private FeatureViewEngine featureViewEngine;
+
+  public FeatureStore() {
```
kennethmhc, line 15: General question: what methods should be included in the java client?

java/hsfs/src/main/java/com/logicalclocks/hsfs/FeatureViewBase.java
```diff
@@ -323,95 +321,6 @@ public List<List<Object>> getFeatureVectors(Map<String, List<Object>> entry, boo
...
-   * Add name/value tag to the feature view.
-   *
-   * @param name
-   *     name of the tag
-   * @param value
-   *     value of the tag. The value of a tag can be any valid json - primitives, arrays or json objects
-   * @throws FeatureStoreException
-   * @throws IOException
-   */
-  public abstract void addTag(String name, Object value) throws FeatureStoreException, IOException;
```
kennethmhc, line 336: why removing all the abstract method? Are they not needed for spark/flink etc? same for `featurestorebase` and `featuregroupbase`.
I think you can implement the abstract method as "not available" in the `FeatureView`

kennethmhc:
When I was testing it, I found a bug. Can you change line 215 in `VectorServer` to
```
      String zippedTupleString =
          zipArraysToTupleString(preparedStatementParameters.get(fgId)
                  .entrySet()
                  .stream()
                  .sorted(Comparator.comparingInt(Map.Entry::getValue))
                  .map(e -> entry.get(e.getKey()))
              .collect(Collectors.toList()));
```
Basically, the problem is that when there are multiple primary key, they need to be sorted according to the index of `preparedStatementParameters`

java/hsfs/src/main/java/com/logicalclocks/hsfs/engine/FeatureViewEngineBase.java
davitbzh: FeatureGroupBaseForApi was used as a dummy class for API calls. In hsfs only abstract classes were present and it is not initializable. Since we added now support for  StreamFeatureGroup, it can do the job

java/hsfs/src/main/java/com/logicalclocks/hsfs/FeatureStore.java
davitbzh: Only thing in java client can be done is fetch FG, FV metadata and get feature vectors. I don't see any other use cases

java/hsfs/src/main/java/com/logicalclocks/hsfs/FeatureViewBase.java
davitbzh: My point was that  who will use `addTag` from Flink or Beam? It doesn't make sense. But it is possible do add since only api classes are used

PR #1115, opened 2023-09-12
[FSTORE-1036] Add usage logging

kennethmhc:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


python/hsfs/usage.py
```diff
@@ -0,0 +1,255 @@
...
+    if not _is_enabled:
+        return func
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        # Disable usage AFTER import hsfs, return function itself
+        if not _is_enabled:
+            return func(*args, **kwargs)
+
+        start_time = time.time()
```
javierdlrm, line 199: In Python 3.3, perf_counter() was introduced which is a bit more accurate ([docs](https://docs.python.org/3/library/time.html#time.perf_counter)). Can we use that one instead?

python/hsfs/usage.py
```diff
@@ -0,0 +1,255 @@
...
+        except Exception as e:
+            exception = e
+            raise e
+        finally:
+            try:
+                end_time = time.time()
+                execution_time = end_time - start_time
+                _method_counter.add(func)
+                # Send log to REST API server
+                if exception or _method_counter.should_sample(func):
```
javierdlrm, line 214: If I understand correctly, the first 100 calls to annotated methods will be logged individually (one request to our logging service per method call), right?
Have you thought/ test/ notice possible performance impact after adding the ThreadPoolExecutor?

python/hsfs/usage.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def _get_method_name(self, m):
+        return m.__module__ + m.__name__
+
+
+logging.basicConfig(stream=sys.__stdout__)
+_logger = logging.getLogger(__name__)
+_logger.setLevel(logging.INFO)
+_conn = http.client.HTTPSConnection("usage.hops.works")
+_executor = concurrent.futures.ThreadPoolExecutor(max_workers=2)
```
javierdlrm, line 119: Can we make `max_workers` configurable as well? Maybe through another environment variable?

python/hsfs/usage.py
kennethmhc: > If I understand correctly, the first 100 calls to annotated methods will be logged individually (one request to our logging service per method call), right?

yes

> Have you thought/ test/ notice possible performance impact after adding the ThreadPoolExecutor?

When logging is off, there is no apparent change in the latency. When it is on, there is around 4ms increase in the latency. In production setting, I would expect user to turn off the usage logging.

javierdlrm: Alright, then we just need to make sure it's well documented especially for get_feature_vector(s) method.
We could highlight it in the docs with one of those `info` captions

javierdlrm (approved):

robzor92 (approved):

robzor92 (approved):

javierdlrm (approved):

PR #1116, opened 2023-09-13
[FSTORE-862][APPEND] Don't provide the current offsets by default when starting job

bubriks:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/engine/python.py
+++ b/python/hsfs/engine/python.py
@@ -1036,7 +1036,7 @@ class Engine:
         elif not isinstance(
             feature_group, ExternalFeatureGroup
         ) and self._start_offline_materialization(offline_write_options):
-            if offline_write_options.get("skip_offsets", False):
+            if offline_write_options.get("skip_offsets", True):
                 # don't provide the current offsets (read from where the job last left off)
                 initial_check_point = ""
             # provide the initial_check_point as it will reduce the read amplification of materialization job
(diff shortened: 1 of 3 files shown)


SirOibaf:
Are you sure this PR is correct? Looks to me that we always end up in the first branch of the if statement as the `initial_check_point` is always not empty string as we set it here:
https://github.com/logicalclocks/feature-store-api/blame/25cfcd57ad792a3b6a732570943692c49b406fbc/python/hsfs/engine/python.py#L956

It doens't seem there is a way of controlling the `skip_offset` parameter in the first branch and the offsets are always skipped.

Additionally the `skip_offset` parameter is not documented anywhere in the APIs. Please add the proper documentation in the insert method.

bubriks:
I think everything should be correct.

`initial_check_point` will be empty if topic doesn't exists (for example after upgrade).

In the first if statement (here: https://github.com/logicalclocks/feature-store-api/blame/25cfcd57ad792a3b6a732570943692c49b406fbc/python/hsfs/engine/python.py#L1016) we always run materialization job setting the initial offset to 0 for all partitions of topic since the topic didn't exist (done here: https://github.com/logicalclocks/feature-store-api/blame/25cfcd57ad792a3b6a732570943692c49b406fbc/python/hsfs/engine/python.py#L1028)

bubriks:
@SirOibaf I also change the `skip_offset` parameter name to `use_current_offsets` as i think its more descriptive.

PR #1134, opened 2023-10-18
[FSTORE-916] FeatureGroup.insert does not retry in case of a failure

bubriks:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/feature_group.py
+++ b/python/hsfs/feature_group.py
@@ -2606,10 +2606,8 @@ class FeatureGroup(FeatureGroupBase):
                     except RestAPIError as e:
                         if e.response.status_code == 404:
                             if e.response.json().get("errorCode", "") == 130009:
-                                print(1111)
                                 break  # no need to retry, since no such job exists
                             else:
-                                print(2222)
                                 time.sleep(1)  # backoff and then retry
                                 continue
                         raise e
--- a/python/hsfs/feature_group.py
+++ b/python/hsfs/feature_group.py
@@ -15,6 +15,8 @@
 #
 
 import copy
+import time
+
 from hsfs.ge_validation_result import ValidationResult
 import humps
 import json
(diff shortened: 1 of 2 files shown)


python/hsfs/feature_group.py
```diff
@@ -2590,23 +2592,26 @@ def parents(self):
...
-
-        return self._materialization_job
+        if self._materialization_job is not None:
+            return self._materialization_job
+        else:
+            feature_group_name = util.feature_group_name(self)
+            job_suffix_list = ["materialization", "backfill"]
+            for job_suffix in job_suffix_list:
+                job_name = "{}_offline_fg_{}".format(feature_group_name, job_suffix)
+                for _ in range(3):  # retry starting job
```
kennethmhc, line 2602: I think the loop order should be reversed.

python/hsfs/feature_group.py
```diff
@@ -2590,23 +2592,26 @@ def parents(self):
...
+                    try:
+                        self._materialization_job = job_api.JobApi().get(job_name)
+                        return self._materialization_job
+                    except RestAPIError as e:
+                        if e.response.status_code == 404:
+                            if e.response.json().get("errorCode", "") == 130009:
+                                break  # no need to retry, since no such job exists
+                            else:
+                                time.sleep(1)  # backoff and then retry
```
kennethmhc, line 2611: sleep should be after each retry but not each retry and suffix.

python/hsfs/feature_group.py
bubriks: In this approach we can confirm that "materialization" or "backfill" job doesn't exists before retrying to get the other one.  If we get errorCode `130009` then there is no need for more retries (if we change the order of loops then we have to implement a solution of remembering which one is confirmed to not exist).

python/hsfs/feature_group.py
bubriks: Like mentioned in the 1st comment it would complicate this method more than necessary in my opinion. Furthermore, for situation where we have to sleep once for all suffixes and then retry, it is highly likely that all requests will fail either way due to service not being up.

kennethmhc (approved):
LGTM!

PR #1139, opened 2023-10-31
[FSTORE-1047] Support Similarity Search in the Feature Store

kennethmhc:
cache required sk

embedding fg right join

fv get vector

fv find neighbour

write to single index

read embedding metadata

read embedding

create embedding

This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


bubriks (changes requested):

robzor92:
I'm not sure how close this is to being merged, but there should also be a PR with workflow tests being merged along with it.

python/hsfs/feature_view.py
```diff
@@ -515,9 +529,73 @@ def get_feature_vectors(
...
+            passed_features=passed_features,
+            external=True,
+            allow_missing=True
+        )
+
+    def _extract_primary_key(self, result_key):
+        primary_key_map = {}
+        for prefix_sk, sk in self._prefix_serving_key_map.items():
+            if prefix_sk in result_key:
+                primary_key_map[sk.required_serving_key] = result_key[prefix_sk]
```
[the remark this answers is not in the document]
kennethmhc: In this case, it needs look up twice anyway because result of `prefix_sk in result_key` and `sk.feature_name in result_key` needs to be saved beforehand.

python/hsfs/core/vector_db_client.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def _rewrite_result_key(self, result_map, key_map):
+        new_map = {}
+        for key in result_map.keys():
+            new_key = key_map.get(key)
+            if new_key is None:
+                raise FeatureStoreException(
+                    f"Feature '{key}' from embedding feature group is not found in the query"
+                )
+            new_map[new_key] = result_map[key]
```
bubriks, line 147:
```suggestion
        for key, value in result_map.items():
            new_key = key_map.get(key)
            if new_key is None:
                raise FeatureStoreException(
                    f"Feature '{key}' from embedding feature group is not found in the query"
                )
            new_map[new_key] = value
```

robzor92 (changes requested):
I saw we started merging the PRs for hopsworks-ee and hopsworks-chef. I really must stress the need for writing workflow tests before we start merging new functionality. It makes the code review easier by getting a concrete example how the code is supposed to work in a complete example, and tests that it works in the first place.

robzor92:
> @robzor92 This is the loadtest: [logicalclocks/loadtest#243](https://github.com/logicalclocks/loadtest/pull/243) I have tested it on VM

Thanks, I'll have a look at the test

python/hsfs/core/opensearch.py
```diff
@@ -0,0 +1,79 @@
...
+        if not cls._instance:
+            cls._instance = super(OpenSearchClientSingleton, cls).__new__(cls)
+            cls._instance._opensearch_client = None
+            cls._instance._setup_opensearch_client()
+        return cls._instance
+
+    def _setup_opensearch_client(self):
+        if not self._opensearch_client:
+            try:
+                import hopsworks
```
robzor92, line 36: We should not use hopsworks library inside hsfs here. Especially since I see we do hopsworks.login() and this is guaranteed to create issues given the complexity of the connection code between hopsworks/hsfs/hsml. This also means the opensearch code being used here needs to be moved to hsfs.

robzor92 (changes requested):

python/hsfs/core/opensearch.py
kennethmhc: isn't the purpose of hopsworks-api is to gather all the common code e.g opensearch, kafka? if so then here it needs to import hopsworks. Not sure what issues will be created regarding the login, but here it can just throw an error and ask the users to login hopsworks by themselves, but essential what they will do is the code here.

robzor92: The purpose of hopsworks is that it will eventually absorb hsfs and hsml. It should not be used from hsfs or hsml directly. It currently provides the functionality that are not in hsfs/hsml because it provides APIs to general services like jobs, kafka, python etc.

robzor92: Yes until we have a solution to manage this. @javierdlrm was working on a design doc at some point for this if you want more info.

kennethmhc: ok then I will duplicate the opensearch-api related code to hsfs.

kennethmhc: @robzor92 removed the hopsworks dependency: https://github.com/logicalclocks/feature-store-api/pull/1139/commits/221dad3506f817f508f6950d953f68024207f345

python/hsfs/core/opensearch.py
```diff
@@ -0,0 +1,70 @@
...
+    def search(self, index=None, body=None):
+        try:
+            return self._opensearch_client.search(body=body, index=index)
+        except self.OpenSearchConnectionError:
+            self._refresh_opensearch_connection()
+            return self._opensearch_client.search(body=body, index=index)
+
+    def close(self):
+        if self._opensearch_client:
+            self._opensearch_client.close()
```
bubriks, line 70:
```suggestion
            self._opensearch_client.close()
            self._opensearch_client = None
```

in `_refresh_opensearch_connection`:
```
self.close()
self._setup_opensearch_client()
```

in `__new__`: (no point of `cls._instance._opensearch_client = None`)
```
if not cls._instance:
    cls._instance = super(OpenSearchClientSingleton, cls).__new__(cls)
    cls._instance._setup_opensearch_client()
```

bubriks (approved):
just minor feedback, but ok either way

python/hsfs/core/opensearch_api.py
```diff
@@ -0,0 +1,136 @@
...
+        self,
+        project_id,
+        project_name,
+    ):
+        self._project_id = project_id
+        self._project_name = project_name
+        self._variable_api = VariableApi()
+
+    def _get_opensearch_url(self):
+        url = os.environ.get(OPENSEARCH_CONFIG.ELASTIC_ENDPOINT_ENV_VAR)
```
robzor92, line 56: We should not be dependent on this ELASTIC_ENDPOINT environment variable. There may be environments where this is not present and that will break the API. Since we have service discovery (consul) for applications within our private network it should use that.

My understanding is we should use the same logic for determining the opensearch url as we do for flying duck:
https://github.com/logicalclocks/feature-store-api/blob/6ed179af26d1005b72b971b88b28cd4724333646/python/hsfs/core/arrow_flight_client.py#L87

That being said I'm not an expert in this domain but the current logic is not correct. On line 77 you do `return f"https://{service_discovery_domain}:9200"` which effectively just becomes `https://consul:9200`?

I'm happy for maybe @SirOibaf which reviewed this corresponding code for flying duck to chime in here.

robzor92 (changes requested):

python/hsfs/core/opensearch_api.py
kennethmhc: We only use it if it is available: https://github.com/kennethmhc/feature-store-api/blob/FSTORE-1047/python/hsfs/core/opensearch_api.py#L57 otherwise we infer the url end point using the flying duck logic: https://github.com/kennethmhc/feature-store-api/blob/FSTORE-1047/python/hsfs/core/opensearch_api.py#L60 (mostly copy from https://github.com/logicalclocks/feature-store-api/blob/master/python/hsfs/core/arrow_flight_client.py#L87)

robzor92: Yes I understand that but the code is more clean and consistent if we have the same logic for determining the endpoint for different services. Checking that environment variable is added complexity in my view and for anyone that didn't write the code.

I saw it was a copy, but for example this line, https://github.com/logicalclocks/feature-store-api/blob/60bfb1a74811b323ed014afa40fc9468e5d23053/python/hsfs/core/arrow_flight_client.py#L102, as I mentioned before I still don't understand how that can work? From my perspective it should be something like:

`host_url = f"https://elastic.service.{service_discovery_domain}:9200"`

kennethmhc: right it should be `f"https://elastic.service.{service_discovery_domain}:9200"`. I think it is necessary to pass the endpoint from environment variable as well, as it should also work with external elastic cluster whose url is different from the hopsworks backend.

robzor92 (approved):
After a chat with @jimdowling it is acceptable if similarity search only works internally in the platform for now. 

I would still like a review from @SirOibaf regards to the connection logic here for external clients: https://github.com/logicalclocks/feature-store-api/blob/36147d24ea30eb751048cab40f9be649ff294523/python/hsfs/core/opensearch_api.py#L55

But I will not block the merging of this PR because of that given the need to demo this to prospects etc.

python/hsfs/core/opensearch_api.py
SirOibaf: The environment variable is not necessary, if you are external to the cluster, then we need to setup the load balancer, if you are inside the cluster, we can use the service discovery and it doesn't matter where elastic is running, it will give back the right IP.

Additionally using the service discovery allows us to handle cases where one elastic node goes down. the client will do the resolution again and consul will give back another IP.

Btw the full name of the elastic dns is `rest.elastic.service.`

kennethmhc: @robzor92 @SirOibaf Removed the environment variable and change the elastic dns to `rest.elastic.service`
https://github.com/logicalclocks/feature-store-api/pull/1139/commits/85300703999f2f567700f57efadf0d25f73d3309

PR #1143, opened 2023-11-03
[FSTORE-1072] Configure timeout for arrow flight

steffengr:
This PR allows to configure the timeout for arrow flight which currently defaults to 30 seconds which is too short. The server has a queue timeout of 10 minutes which means the default should be longer than 10 minutes. Set it to 15 minutes and allow the user to configure it so that long running queries can be supported.

JIRA Issue: https://hopsworks.atlassian.net/browse/FSTORE-1072

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/core/arrow_flight_client.py
+++ b/python/hsfs/core/arrow_flight_client.py
@@ -54,6 +54,7 @@ class ArrowFlightClient:
     FILTER_NUMERIC_TYPES = ["bigint", "tinyint", "smallint", "int", "float", "double"]
     READ_ERROR = 'Could not read data using ArrowFlight. If the issue persists, use read_options={"use_hive": True} instead.'
     WRITE_ERROR = 'Could not write data using ArrowFlight. If the issue persists, use write_options={"use_spark": True} instead.'
+    DEFAULT_TIMEOUT = 900
 
     def __init__(self):
         try:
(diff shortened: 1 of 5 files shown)


python/hsfs/core/arrow_flight_client.py
```diff
@@ -54,6 +54,7 @@ class ArrowFlightClient:
     FILTER_NUMERIC_TYPES = ["bigint", "tinyint", "smallint", "int", "float", "double"]
     READ_ERROR = 'Could not read data using ArrowFlight. If the issue persists, use read_options={"use_hive": True} instead.'
     WRITE_ERROR = 'Could not write data using ArrowFlight. If the issue persists, use write_options={"use_spark": True} instead.'
+    DEFAULT_TIMEOUT = 900
```
kennethmhc, line 57: > The server has a queue timeout of 10 minutes by default and the client should match that configuration.
should this be 600?

steffengr: No. I added 300 seconds to the 600 seconds maximum queue time that the server has configured as it might hang for 600 seconds in the worst case and should then have some head room to execute.

kennethmhc (approved):
LGTM

PR #1168, opened 2023-12-15
[FSTORE-1116] support aiomysql for executing parallel online queries

DhananjayMukhedkar:
This PR adds/fixes/changes...

- adding support for  aiomysql engine.  This replaces sqlalchemy engine during get_feature_vector(s). Connection pool size params can be set by user as `options` in `init_serving`.
- the code to construct the results from the query is the same as before
- tested with existing loadtest on vm

JIRA Issue: -
https://hopsworks.atlassian.net/browse/FSTORE-1116

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


python/hsfs/feature_view.py
```diff
@@ -193,6 +193,7 @@ def init_serving(
         training_dataset_version: Optional[int] = None,
         external: Optional[bool] = None,
         options: Optional[dict] = None,
+        parallel=False,
```
jimdowling, line 196: Why is it default parallel "False"?

python/hsfs/core/vector_server.py
```diff
@@ -658,6 +697,204 @@ def filter_entry_by_join_index(self, entry, join_index):
...
+            # add partial results to the global results
+            for i, entry in enumerate(entries):
+                batch_results[i].update(
+                    statement_results.get(
+                        self._get_result_key_serving_key(serving_keys, entry), {}
+                    )
+                )
+        return batch_results, serving_keys
+
+    async def _set_aiomysql_connection(self, options):
```
jimdowling, line 832: Can you add an easy parameter for users to specify the level of parallelism (number of MySQL connections)?
The online_connector does not have the 'minsize' and 'maxsize' parameters for the connection pool size:

 pool = await aiomysql.create_pool(host=hostname, port=3306,
                                   user=uname, password=pwd, db=dbname, loop=loop, 
                                   minsize=4, maxsize=10)

python/hsfs/feature_view.py
jimdowling: Should we not just have the number of MySQL Connections (threads) instead of a boolean?
I would prefer it to be discovered automatically - and you can override it if needed, rather than set at '1' by default.
For example, Polars doesn't give you one thread, it discovers the number of cores on your machine and sets it to that by default.

The level of parallelism we need is constrained by:
1. how many parallel lookups there will be
2. how many available MySQL connections we can use

python/hsfs/core/vector_server.py
```diff
@@ -82,17 +84,29 @@ def __init__(
...
 
     def init_serving(
-        self, entity, batch, external, inference_helper_columns=False, options=None
+        self,
+        entity,
+        batch,
+        external,
+        inference_helper_columns=False,
+        options=None,
+        parallel=False,
```
jimdowling, line 96: parallel should be a number, not boolean flag.
And it should be 'none'. By default, we should have it as the number of parallel threads (MySQL Connections) - the number of prepared statements.

Most often, a hsfs client is used by one model and every call will have the same number of prepared statements. We can set the minsize and maxsize for aiomysql.create_pool(..) to be the same value. That will ensure it maximizes performance.

jimdowling (changes requested):
We should allow users to explicitly set the level of parallelism, and we should default set parallelism to the number of prepared statements.

python/hsfs/core/vector_server.py
DhananjayMukhedkar: minsize and maxsize arguments can be set by users in dict `options` parameter. e.g. `init_serving(options={'minsize':1,'maxsize':1})` which is consistent with existing API

python/hsfs/feature_view.py
DhananjayMukhedkar: fixed default pool size to now use the number of prepared statements as minsize. The user can still override it by setting the minsize argument in `options` dict in `init_serving`.

python/hsfs/core/vector_server.py
DhananjayMukhedkar: Removed the parallel flag and the sqlalchemy engine entirely during get_feature_vector(s)

jimdowling (approved):

python/hsfs/core/vector_server.py
```diff
@@ -383,114 +405,124 @@ def _vector_result(self, entry, prepared_statement_objects):
...
-                for row in result_proxy:
-                    result_dict = self.deserialize_complex_features(
-                        self._complex_features, row._asdict()
-                    )
-                    serving_vector.update(result_dict)
+        # run all the prepared statements in parallel using aiomysql engine
+        loop = asyncio.get_event_loop()
+        results = loop.run_until_complete(
+            self._execute_prep_statements(
+                list(prepared_statement_objects.values()), bind_entries
```
kennethmhc, line 433: does it guarantee that the order of `prepared_statement_objects.values()` and `bind_entries` are the same?

python/hsfs/core/vector_server.py
```diff
@@ -383,114 +405,124 @@ def _vector_result(self, entry, prepared_statement_objects):
...
+            statement_results = {}
+            serving_keys = self._serving_key_by_serving_index[prepared_statement_index]
+            # Use prefix from prepare statement because prefix from serving key is collision adjusted.
+            prefix_features = [
+                (self._prefix_by_serving_index[prepared_statement_index] or "")
+                + sk.feature_name
+                for sk in self._serving_key_by_serving_index[prepared_statement_index]
+            ]
+            # iterate over results by index of the prepared statement
+            for row in parallel_results[count]:
```
kennethmhc, line 506: length of `prepared_statement_objects` can be longer than that of `parallel_results` because some prepare statement may not be executed.
```
            if prepared_statement_index not in self._serving_key_by_serving_index:
                continue
```

python/hsfs/core/vector_server.py
robzor92: @dhananjay-mk would be good to look at the comments here

PR #1174, opened 2023-12-21
[FSTORE-1130] Get schema from shared project

bubriks:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


kennethmhc (approved):
LGTM!

SirOibaf (changes requested):
Can you cleanup the code here since it's not needed anymore: https://github.com/logicalclocks/feature-store-api/blob/cfb6598627be54fefce8244ba72f9f170121eb19/python/hsfs/feature_view.py#L2817

Also, that cache is useless, since the method call in the default value of the `fs_cache.get()` call is always evaluated before getting the value from the map. This means that for each feature we are calling the APIs to retrieve the feature store object. Which is expensive for no reason.

SirOibaf (approved):

PR #1175, opened 2023-12-21
[FSTORE-1118] Pandas arrow type dataframes cannot be inserted into feature group

manu-sj:
This PR fixes issue of pyarrow types introduced as part of pandas 2.0 not being insterted into the feature store.

- Changes were made in this PR adds support to convert for arrow datatype introduced in pandas 2.0.
- This PR changes the way the conversion between pandas datatypes were performed to increase readability and maintainability of the code. Dicussion done in design log - https://hopsworks.atlassian.net/wiki/spaces/FST/pages/279314450/Conversion+of+Pandas+arrow+datatypes+to+hopsworks+internal+Datatypes
- Testing involved checking if all types and corresponding arrow types mentioned as part of hopsworks documentation is inserted into the feature group.
- Some datatypes present introducted as part of pyarrow in pandas are not added since they were not explicitly mentioned in the hopsworks documentation. The datatypes not added are mentioned are - uint64, large_binary, large_string, large_utf8, large_list, fixed_size_list (pyarrow list with max size mentioned), maps, dictonary (used for efficent array storage in pyarrow)  

JIRA Issue: - https://hopsworks.atlassian.net/browse/FSTORE-1118

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [x] Unit Tests
- [ ] Integration Tests
- [x] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/engine/python.py
+++ b/python/hsfs/engine/python.py
@@ -92,7 +92,10 @@ PYARROW_HOPSWORKS_DTYPE_MAPPING = {
     **dict.fromkeys([pa.float16(), pa.float32()], "float"),
     **dict.fromkeys([pa.float64()], "double"),
     **dict.fromkeys(
-        [pa.timestamp("ns")] + [pa.timestamp("ns", tz=tz) for tz in pytz.all_timezones],
+        [pa.timestamp("ns")] + [pa.timestamp("ns", tz=tz) for tz in pytz.all_timezones] +
+        [pa.timestamp("us")] + [pa.timestamp("us", tz=tz) for tz in pytz.all_timezones] +
+        [pa.timestamp("ms")] + [pa.timestamp("ms", tz=tz) for tz in pytz.all_timezones] +
+        [pa.timestamp("s")] + [pa.timestamp("s", tz=tz) for tz in pytz.all_timezones],
         "timestamp",
     ),
     **dict.fromkeys([pa.bool_()], "boolean"),
--- a/python/hsfs/engine/python.py
+++ b/python/hsfs/engine/python.py
@@ -31,6 +31,7 @@ import decimal
 import numbers
 import math
 import os
+import pytz
 from datetime import datetime, timezone
 
 import great_expectations as ge
@@ -468,11 +469,11 @@ class Engine:
     def parse_schema_feature_group(self, dataframe, time_travel_format=None):
         arrow_schema = pa.Schema.from_pandas(dataframe)
         features = []
-        for feat_name, feat_type in dataframe.dtypes.items():
+        for feat_name, _ in dataframe.dtypes.items():
             name = feat_name.lower()
             try:
                 converted_type = self._convert_pandas_dtype_to_offline_type(
-                    feat_type, arrow_schema.field(feat_name).type
+                    arrow_schema.field(feat_name).type
                 )
             except ValueError as e:
                 raise FeatureStoreException(f"Feature '{name}': {str(e)}")
@@ -1154,14 +1155,17 @@ class Engine:
         return config
 
     @staticmethod
-    def _convert_pandas_dtype_to_offline_type(dtype, arrow_type):
+    def _convert_pandas_dtype_to_offline_type(arrow_type):
         # This is a simple type conversion between pandas dtypes and pyspark (hive) types,
-        # using pyarrow types to convert "O (object)"-typed fields.
+        # using pyarrow types obatined from pandas dataframe to convert pandas typed fields,
+        # A recurisive function  "_convert_pandas_object_type_to_offline_type" is used to convert complex types like lists and structures
+        # "_convert_simple_pandas_dtype_to_offline_type" is used to convert simple types
         # In the backend, the types specified here will also be used for mapping to Avro types.
-        if dtype == np.dtype("O"):
+
+        if pa.types.is_list(arrow_type) or pa.types.is_struct(arrow_type):
             return Engine._convert_pandas_object_type_to_offline_type(arrow_type)
 
-        return Engine._convert_simple_pandas_dtype_to_offline_type(dtype)
+        return Engine._convert_simple_pandas_dtype_to_offline_type(arrow_type)
 
     @staticmethod
     def convert_spark_type_to_offline_type(spark_type_string):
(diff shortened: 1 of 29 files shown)


python/hsfs/engine/python.py
```diff
@@ -468,11 +469,11 @@ def convert_to_default_dataframe(self, dataframe):
     def parse_schema_feature_group(self, dataframe, time_travel_format=None):
         arrow_schema = pa.Schema.from_pandas(dataframe)
         features = []
-        for feat_name, feat_type in dataframe.dtypes.items():
+        for feat_name, _ in dataframe.dtypes.items():
```
bubriks, line 472:
```suggestion
        for feat_name in arrow_schema.names:
```

bubriks (changes requested):

python/tests/engine/test_python.py
```diff
@@ -1102,7 +1105,7 @@ def test_convert_simple_pandas_type_uint8(self):
 
         # Act
         result = python_engine._convert_simple_pandas_dtype_to_offline_type(
-            dtype=np.dtype("uint8")
+            arrow_type=pa.from_numpy_dtype(np.dtype("uint8"))
```
bubriks, line 1108: Not a big deal but instead of doing `pa.from_numpy_dtype(np.dtype("uint8"))` it would be cleaner to do `pa.uint8()` (this idea applies in multiple places in `test_python.py`)

python/tests/engine/test_python.py
```diff
@@ -1567,6 +1572,346 @@ def test_infer_type_pyarrow_struct_with_list_of_struct_fields(self):
...
+        pdf = pd.DataFrame({"data": pd.Series([1, 5], dtype="int32[pyarrow]")})
+        arrow_schema = pa.Schema.from_pandas(pdf)
+
+        python_engine = python.Engine()
+
+        # Act
+        arrow_type = python_engine._convert_simple_pandas_dtype_to_offline_type(
+            arrow_schema.field("data").type
+        )
+        print(arrow_type)
```
bubriks, line 1618: `print` should be removed (multiple places)

python/tests/engine/test_python.py
```diff
@@ -1567,6 +1572,346 @@ def test_infer_type_pyarrow_struct_with_list_of_struct_fields(self):
...
+        # Arrange
+        pdf = pd.DataFrame({"data": pd.Series([1, 5], dtype="int32[pyarrow]")})
+        arrow_schema = pa.Schema.from_pandas(pdf)
+
+        python_engine = python.Engine()
+
+        # Act
+        arrow_type = python_engine._convert_simple_pandas_dtype_to_offline_type(
+            arrow_schema.field("data").type
+        )
```
bubriks, line 1617:
```suggestion
        python_engine = python.Engine()

        # Act
        arrow_type = python_engine._convert_simple_pandas_dtype_to_offline_type(
            pa.int32()
        )
```

bubriks: could be done like this in multiple tests, instead of creating data frame.

python/tests/engine/test_python.py
```diff
@@ -1567,6 +1572,346 @@ def test_infer_type_pyarrow_struct_with_list_of_struct_fields(self):
...
+        print(arrow_type)
+
+        # Assert
+        assert arrow_type == "int"
+
+    def test_infer_type_pandas_pyarrow_int32(self):
```
bubriks, line 1607: it looks like these tests are duplicates, for example we already have this case covered: https://github.com/logicalclocks/feature-store-api/pull/1175/files#diff-01253ce273d2953406cf0ee612bf433ed4fdaae72f49126492cfe23841beb082L1147

Make sure there are no duplicates. And our naming for test methods is `test_[testedMethodName]_[condition]`. In this case it should be something like`test_convert_simple_pandas_dtype_to_offline_type_int32`

bubriks (changes requested):

python/tests/engine/test_python.py
```diff
@@ -1088,12 +1088,13 @@ def test_convert_pandas_type_object(self, mocker):
         python_engine = python.Engine()
 
         # Act
+        python_engine._convert_pandas_dtype_to_offline_type(arrow_type=pa.struct({}))
         python_engine._convert_pandas_dtype_to_offline_type(
-            dtype=np.dtype("O"), arrow_type=None
+            arrow_type=pa.list_(pa.int32())
         )
```
bubriks, line 1094: should be removed.

python/tests/engine/test_python.py
```diff
@@ -1088,12 +1088,13 @@ def test_convert_pandas_type_object(self, mocker):
...
 
         # Assert
-        assert mock_python_engine_infer_type_pyarrow.call_count == 1
+        assert mock_python_engine_infer_type_pyarrow.call_count == 2
```
bubriks, line 1097:
```suggestion
        assert mock_python_engine_infer_type_pyarrow.call_count == 1
```

python/tests/engine/test_python.py
```diff
@@ -1088,12 +1088,13 @@ def test_convert_pandas_type_object(self, mocker):
         python_engine = python.Engine()
 
         # Act
+        python_engine._convert_pandas_dtype_to_offline_type(arrow_type=pa.struct({}))
```
bubriks, line 1091: we should check the results of convert for `struct` and `list` types, the same way as for `int64` for example.

bubriks: the patching can be removed, the call count is not as important as the result. The same applies in other tests as well, i think it was a mistake that we patched as much as we did so it should be fixed.

python/tests/engine/test_python.py
bubriks: this test is for `struct` not `list`

python/tests/engine/test_python.py
bubriks: so for example for list it would look like this:
```
    def test_convert_pandas_object_type_to_offline_type_list(self, mocker):
        # Arrange
        python_engine = python.Engine()

        # Act
        result = python_engine._convert_pandas_object_type_to_offline_type(
            arrow_type=pa.list_(pa.int8())
        )

        # Assert
        assert result == "array<int>"
```

bubriks (changes requested):

python/tests/engine/test_python.py
manu-sj: I removed the entire function `test_convert_pandas_type_object` since there was already another function - `test_infer_type_pyarrow_struct` that does the test for an empty sturctures. Similarly `test_convert_pandas_type` was also removed since there was another function `test_convert_simple_pandas_type_int64` that does the same test when the call count is removed.

python/tests/engine/test_python.py
```diff
@@ -1258,39 +1236,30 @@ def test_convert_simple_pandas_type_other(self):
...
         # Act
         result = python_engine._convert_pandas_object_type_to_offline_type(
             arrow_type=pa.list_(pa.int8())
         )
 
         # Assert
-        assert result == "array<test>"
+        assert result == "array<int>"
 
     def test_infer_type_pyarrow_struct(self, mocker):
```
bubriks, line 1259: one minor improvement for this test:
```
    def test_infer_type_pyarrow_struct(self):
        # Arrange
        python_engine = python.Engine()

        # Act
        result = python_engine._convert_pandas_object_type_to_offline_type(
            arrow_type=pa.struct([pa.field('f1', pa.int32())])
        )

        # Assert
        assert result == "struct<f1:int>"
```

bubriks: also optionally you can remove `mocker` args from test methods where they are not needed. (but ok if you wish to not do it :) )

python/hsfs/engine/python.py
```diff
@@ -82,6 +83,31 @@
...
+    **dict.fromkeys([pa.bool_()], "boolean"),
+    **dict.fromkeys(
+        [
+            pa.string(),
+            pa.dictionary(value_type=pa.string(), index_type=pa.int8(), ordered=True),
+            pa.dictionary(value_type=pa.string(), index_type=pa.int8(), ordered=False),
+        ],
+        "string",
+    ),  # Category type in pandas stored as dictinoary in pyarrow
+    **dict.fromkeys([pa.date32(), pa.date64()], "date"),
```
bubriks, line 107: `pa.date64()` not in tests (only `date32`)

bubriks (changes requested):

bubriks (approved):

python/hsfs/engine/python.py
```diff
@@ -82,6 +83,31 @@
...
+# Decimal types are currently not supported
+PYARROW_HOPSWORKS_DTYPE_MAPPING = {
+    **dict.fromkeys(
+        [pa.uint8(), pa.uint16(), pa.int8(), pa.int16(), pa.int32()], "int"
+    ),
+    **dict.fromkeys([pa.uint32(), pa.int64()], "bigint"),
+    **dict.fromkeys([pa.float16(), pa.float32()], "float"),
+    **dict.fromkeys([pa.float64()], "double"),
+    **dict.fromkeys(
+        [pa.timestamp("ns")] + [pa.timestamp("ns", tz=tz) for tz in pytz.all_timezones],
```
bubriks, line 95: this will have to be updated after this is merged: https://github.com/logicalclocks/feature-store-api/pull/1169/files

bubriks (changes requested):
this will have to be updated after this is merged: https://github.com/logicalclocks/feature-store-api/pull/1169/files

bubriks (approved):
Looks good

PR #1181, opened 2024-01-02
[FSTORE-1034] .select() method should not default to empty list

bubriks:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


SirOibaf (commented):
To be honest, that's not what I was thinking with this JIRA. I see that you are adding some validation in the Query constructor to validate that users have selected features, however this validation is done in the backend already by https://github.com/logicalclocks/hopsworks-ee/pull/1649 - and it should be done in the backend so that it covers all the clients and UI.

The idea of this JIRA was to change this line: https://github.com/logicalclocks/feature-store-api/blob/8458c361e3e8f45d1ecb1fb0ce99326dce0b41a2/python/hsfs/feature_group.py#L240
so that the `select()` method does not default to empty list, which is going to trigger the error you added in the backend PR.

(and have a look at the java code in case we were doing something similar)

PR #1213, opened 2024-02-06
[FSTORE-1183] Error downloading credentials for Kafka storage connector in Python mode

bubriks:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


python/hsfs/util.py
```diff
@@ -133,6 +133,13 @@ def get_host_name():
     return host
 
 
+def get_dataset_type(path: str):
+    if re.match(r"^(?:dfs://|)/apps/hive/warehouse/*", path):
```
SirOibaf, line 137: I really don't understand this regex. A hopsworks path can either start with `/` or `hdfs://` (or `hopsfs://` for that matter), but for sure not with `dfs://` - So I'm really not sure why we are checking if it begins with `dfs://`

If I try the regex in a Python interpreter it doesn't match the following string: `hdfs:///apps/hive/warehouse/test` which is a problem considering that here: https://github.com/logicalclocks/feature-store-api/blob/2cd5b2f7fdcca9976341111417bd03bfa43bb8cd/python/hsfs/engine/python.py#L905
we add `hdfs://` at the beginning of the file if the file doesn't contain any scheme.

I also checked a couple of storage connector, the rest API doesn't return any scheme, meaning `hdfs://` is added by the line above.

This also means, that the default behaviour of uploading the credentials in the storage connector directory within the Hive warehouse directory is now broken with this PR because the regex doesn't match `hdfs://`.

When you test this PR, please test both cases.

SirOibaf (changes requested):

python/hsfs/util.py
bubriks: yeah sorry i am not sure how i lost the `h` from `hdfs` and probably didnt catch it because of this if statement: https://github.com/logicalclocks/feature-store-api/blob/master/python/hsfs/engine/python.py#L908

SirOibaf (approved):

PR #1216, opened 2024-02-07
[FSTORE-1218] Fix as_of and read_changes doc

marcopellegrinoit:
Added banner in as_of and read_changes about the fact that the methods are not available in Python, but only in Spark.

### DIFF
--- a/python/hsfs/feature_group.py
+++ b/python/hsfs/feature_group.py
@@ -2110,8 +2110,8 @@ class FeatureGroup(FeatureGroupBase):
                     `as_of(end_wallclock_time, exclude_until=start_wallclock_time).read(read_options=read_options)`
                     instead.
 
-        !!! warning "Not available in Python"
-            `read_changes` method is available in Spark but not in Python.
+        !!! warning "Available only using Pyspark/Spark"
+            HUDI supports Time Travel and Incremental Query via Spark context, exclusively in PySpark/Spark
 
         This function only works on feature groups with `HUDI` time travel format.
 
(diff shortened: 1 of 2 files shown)


python/hsfs/feature_group.py
```diff
@@ -2110,7 +2110,8 @@ def read_changes(
                     `as_of(end_wallclock_time, exclude_until=start_wallclock_time).read(read_options=read_options)`
                     instead.
 
-        This function only works on feature groups with `HUDI` time travel format.
+        !!! warning "Warning"
+            HUDI supports Time Travel and Incremental Query via Spark context, exclusively in PySpark/Spark
```
davitbzh, line 2114:
```suggestion
            Time Travel and Incremental Queries are supported exclusively in PySpark/Spark.
```

python/hsfs/feature_group.py
```diff
@@ -2847,6 +2848,9 @@ def as_of(
     ):
         """Get Query object to retrieve all features of the group at a point in the past.
 
+        !!! warning
+            HUDI supports Time Travel and Incremental Query via Spark context, exclusively in PySpark/Spark
```
davitbzh, line 2852:
```suggestion
            Time Travel and Incremental Queries are supported exclusively in PySpark/Spark.
```

python/hsfs/feature_group.py
```diff
@@ -2911,9 +2915,6 @@ def as_of(
                 .as_of("2020-10-20", exclude_until="2020-10-19")
             ```
 
-        !!! warning
-            This function only works for feature groups with time_travel_format='HUDI'.
-
```
davitbzh, line 2916: you still need time travel enabled feature group, but we now also support time_travel_format='DELTA'

davitbzh (changes requested):

vatj (approved):

PR #1226, opened 2024-02-17
[FSTORE-1247] Raise DataValidationException on validation failure of a dataframe if the validation ingestion policy is set to strict

vatj:
Added for both regular and external feature group. Added docs for external feature group insert method. 

Note that insert method of external feature group needs to be "fixed", as it is confusing. There is no materialization job or need to wait for example. To be addressed in separate PR

JIRA Issue: -

Priority for Review: - 

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


python/hsfs/core/external_feature_group_engine.py
```diff
@@ -80,7 +80,12 @@ def insert(
         )
 
         if ge_report is not None and ge_report.ingestion_result == "REJECTED":
-            return None, ge_report
+            raise DataValidationException(
+                "Data validation failed while validation ingestion policy set to strict, "
+                + f"insertion to {feature_group.name} was aborted.\n"
+                + f"Check a summary of your report at {self._get_feature_group_url(feature_group)}, "
+                + f"or download the full report from {ge_report._full_report_path}."
```
SirOibaf, line 87: Seems to me that the `ge_report._full_report_path` returns the path on the file system - so this error seems a bit vague as the user won't be able to click on it to download the report. 
Maybe we should redirect the user to the feature group overview UI where they can see the result?

python/hsfs/core/external_feature_group_engine.py
```diff
@@ -134,3 +139,14 @@ def update_deprecated(self, feature_group, deprecate):
         self._feature_group_api.update_metadata(
             feature_group, copy_feature_group, "deprecate", deprecate
         )
+
+    def _get_feature_group_url(self, feature_group):
```
SirOibaf, line 143: Minor - Feels like we can put this method in the util module instead of having duplicates here and in the `feature_group_engine.py`

python/hsfs/core/feature_group_engine.py
```diff
@@ -107,7 +107,12 @@ def insert(
         )
 
         if ge_report is not None and ge_report.ingestion_result == "REJECTED":
-            return None, ge_report
+            raise exceptions.DataValidationException(
+                "Data validation failed while validation ingestion policy set to strict, "
+                + f"insertion to {feature_group.name} was aborted.\n"
+                + f"Check a summary of your report at {self._get_feature_group_url(feature_group)}, "
+                + f"or download the full report from {ge_report._full_report_path}."
```
SirOibaf, line 114: Same comment about the url

SirOibaf (changes requested):

python/hsfs/core/external_feature_group_engine.py
vatj: Will refactor it

python/hsfs/core/external_feature_group_engine.py
vatj: Good catch, that does sound more helpful

SirOibaf (approved):

PR #1255, opened 2024-03-18
[FSTORE-1186][APPEND] Polars Integration into Hopsworks - Writing and Reading DataFrames

manu-sj:
This PR fixes two issue present in the integration of Polars to Hopsworks

1. **Helper Columns** - When reading helper columns as a polars dataframe incorrect columns with empty data was obatined.

- _Root Cause_: Issue caused due to incorrect schema being used to create the polars dataframe. Polars dataframe was created using schema extracted from feature vectors.
- _Fix done_ : Corrected the schema used by extracting the inference helper columns.

3. **Training Data** - When creating training data from a feature group with very little data, an exception is encountered sporadically.

- _Root Cause_ : Issue caused due to the very little data being available in the feature group. When very little training data is present in the feature group there is a possibility that one of the training data split (train, test or validation) can be empty. Spark writes an empty file in this case. When reading from polars empty file were not read into a list since it causes an exception to be throw while merging into a single dataframe due to schema mismatch. Hence a empty list would be obtained if the entire split is empty which causes an exception when trying to merge with split.
- _Fix done_ : Check added to see if df list is empty and return an empty polars dataframe with column names in that usecase.

**How Has This Been Tested?**

- [x] Unit Tests
- [ ] Integration Tests
- [x] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/core/vector_server.py
+++ b/python/hsfs/core/vector_server.py
@@ -60,11 +60,6 @@ class VectorServer:
             if features
             else []
         )
-        self._inference_helper_col_name = (
-            [feat.name for feat in features if feat.inference_helper_column]
-            if features
-            else []
-        )
         self._skip_fg_ids = skip_fg_ids or set()
         self._prepared_statement_engine = None
         self._prepared_statements = None
@@ -378,7 +373,7 @@ class VectorServer:
         elif return_type.lower() == "polars":
             # Polar considers one dimensional list as a single columns so passed vector as 2d list
             polars_df = pl.DataFrame(
-                [serving_vector], schema=self._inference_helper_col_name, orient="row"
+                [serving_vector], schema=serving_vector.keys(), orient="row"
             )
             return polars_df
         else:
@@ -416,7 +411,9 @@ class VectorServer:
             return pd.DataFrame(batch_results)
         elif return_type.lower() == "polars":
             polars_df = pl.DataFrame(
-                batch_results, schema=self._inference_helper_col_name, orient="row"
+                batch_results,
+                schema=feature_view_object.inference_helper_columns,
+                orient="row",
             )
             return polars_df
         else:
--- a/python/hsfs/core/vector_server.py
+++ b/python/hsfs/core/vector_server.py
@@ -60,6 +60,11 @@ class VectorServer:
             if features
             else []
         )
+        self._inference_helper_col_name = (
+            [feat.name for feat in features if feat.inference_helper_column]
+            if features
+            else []
+        )
         self._skip_fg_ids = skip_fg_ids or set()
         self._prepared_statement_engine = None
         self._prepared_statements = None
@@ -373,7 +378,7 @@ class VectorServer:
         elif return_type.lower() == "polars":
             # Polar considers one dimensional list as a single columns so passed vector as 2d list
             polars_df = pl.DataFrame(
-                [serving_vector], schema=self._feature_vector_col_name, orient="row"
+                [serving_vector], schema=self._inference_helper_col_name, orient="row"
             )
             return polars_df
         else:
@@ -411,7 +416,7 @@ class VectorServer:
             return pd.DataFrame(batch_results)
         elif return_type.lower() == "polars":
             polars_df = pl.DataFrame(
-                batch_results, schema=self._feature_vector_col_name, orient="row"
+                batch_results, schema=self._inference_helper_col_name, orient="row"
             )
             return polars_df
         else:
--- a/python/hsfs/engine/python.py
+++ b/python/hsfs/engine/python.py
@@ -351,7 +351,7 @@ class Engine:
     # To read the training dataset content, this to avoid the pydoop dependency
     # requirement and allow users to read Hopsworks training dataset from outside
     def _read_hopsfs_remote(
-        self, location, data_format, read_options={}, dataframe_type="default"
+        self, location, data_format, read_options=None, dataframe_type="default"
     ):
         total_count = 10000
         offset = 0
--- a/python/hsfs/engine/python.py
+++ b/python/hsfs/engine/python.py
@@ -274,9 +274,16 @@ class Engine:
                 )
             )
         if dataframe_type.lower() == "polars":
-            return self._return_dataframe_type(
-                pl.concat(df_list), dataframe_type=dataframe_type
-            )
+            # Below check performed since some files materialized when creating training data are empty
+            # If empty dataframe is in df_list then polars cannot concatenate df_list due to schema mismatch
+            # However if the entire split contains only empty files which can occur when the data size is very small then one of the empty dataframe is return so that the column names can be accessed.
+            non_empty_df_list = [df for df in df_list if not df.is_empty()]
+            if non_empty_df_list:
+                return self._return_dataframe_type(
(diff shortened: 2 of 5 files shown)


python/hsfs/engine/python.py
```diff
@@ -321,7 +328,9 @@ def _read_hopsfs(
         try:
             from pydoop import hdfs
         except ModuleNotFoundError:
-            return self._read_hopsfs_remote(location, data_format, read_options or {})
+            return self._read_hopsfs_remote(
+                location, data_format, read_options, dataframe_type
```
vatj, line 332: Why did you remove the `or {}` part?

python/hsfs/engine/python.py
```diff
@@ -364,13 +371,20 @@ def _read_hopsfs_remote(self, location, data_format, read_options=None):
                     ):
                         arrow_flight_config = read_options.get("arrow_flight_config")
                         df = arrow_flight_client.get_instance().read_path(
-                            inode.path, arrow_flight_config
+                            inode.path,
+                            arrow_flight_config,
+                            dataframe_type=dataframe_type,
```
vatj, line 376: Why do you need to pass dataframe type here and use it again below?

python/hsfs/engine/spark.py
```diff
@@ -532,8 +532,12 @@ def get_training_data(
 
     def split_labels(self, df, labels, dataframe_type):
         if labels:
-            labels_df = df.select(*labels)
-            df_new = df.drop(*labels)
+            if isinstance(df, pd.DataFrame):
```
vatj, line 535: Why is this needed? Was it not working before? Or do you need this for polars instead of pandas?

python/tests/engine/test_python.py
```diff
@@ -249,6 +249,68 @@ def test_read_hopsfs_connector(self, mocker):
...
 
+    def test_read_hopsfs_connector_empty_dataframe(self, mocker):
+        # Arrange
+
+        # Setting list of empty dataframes as return value from _read_hopsfs
+        mock_python_engine_read_hopsfs = mocker.patch(
+            "hsfs.engine.python.Engine._read_hopsfs",
+            return_value=[pd.DataFrame(), pd.DataFrame()],
+        )
+        mock_python_engine_read_s3 = mocker.patch("hsfs.engine.python.Engine._read_s3")
```
vatj, line 260: Why do we need to patch S3 here?

python/tests/engine/test_python.py
```diff
@@ -249,6 +249,68 @@ def test_read_hopsfs_connector(self, mocker):
...
+            storage_connector=connector,
+            data_format="csv",
+            read_options=None,
+            location=None,
+            dataframe_type="default",
+        )
+
+        # Assert
+        assert mock_python_engine_read_hopsfs.call_count == 1
+        assert mock_python_engine_read_s3.call_count == 0
```
vatj, line 279: Why do we need to check that we don't read from S3?

python/tests/engine/test_python.py
```diff
@@ -249,6 +249,68 @@ def test_read_hopsfs_connector(self, mocker):
...
+
+    def test_read_hopsfs_connector_empty_dataframe_polars(self, mocker):
+        # Arrange
+
+        # Setting empty list as return value from _read_hopsfs
+        mock_python_engine_read_hopsfs = mocker.patch(
+            "hsfs.engine.python.Engine._read_hopsfs",
+            return_value=[pl.DataFrame(), pl.DataFrame()],
+        )
+        mock_python_engine_read_s3 = mocker.patch("hsfs.engine.python.Engine._read_s3")
```
vatj, line 291: Same?

python/hsfs/engine/python.py
```diff
@@ -274,9 +274,16 @@ def read(
...
+            # Below check performed since some files materialized when creating training data are empty
+            # If empty dataframe is in df_list then polars cannot concatenate df_list due to schema mismatch
+            # However if the entire split contains only empty files which can occur when the data size is very small then one of the empty dataframe is return so that the column names can be accessed.
+            non_empty_df_list = [df for df in df_list if not df.is_empty()]
+            if non_empty_df_list:
+                return self._return_dataframe_type(
+                    pl.concat(non_empty_df_list), dataframe_type=dataframe_type
+                )
+            else:
+                return df_list[0]
```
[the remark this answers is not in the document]
manu-sj: Yes that is the problem. Since there is no data polars cannot infer the type. Pandas use the "object" type in these usecases.

python/hsfs/engine/python.py
manu-sj: Merging mistake sorry corrected.

python/hsfs/engine/python.py
manu-sj: Arrow flight returns the dataframe as pandas by default. I had modifed `_get_dataset` function in _arrow_flight_client.py_ to read as polars dataframe when the dataframe type is polars. So this passing of `dataframe_type` is used to acheive this functionality.

python/tests/engine/test_python.py
manu-sj: Not needed removed.

python/tests/engine/test_python.py
manu-sj: Not needed removed.

python/tests/engine/test_python.py
manu-sj: Not needed removed.

python/hsfs/core/vector_server.py
```diff
@@ -373,7 +378,7 @@ def get_inference_helper(self, entry, return_type):
         elif return_type.lower() == "polars":
             # Polar considers one dimensional list as a single columns so passed vector as 2d list
             polars_df = pl.DataFrame(
-                [serving_vector], schema=self._feature_vector_col_name, orient="row"
+                [serving_vector], schema=self._inference_helper_col_name, orient="row"
```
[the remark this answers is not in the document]
manu-sj: Changed it to use serving_vector.keys() for `get_inference_helper`.

For `get_inference_helpers` since `batch_results` is a list I used `feature_view_object.inference_helper_columns` which was a pre-existing variable in feature view. I am wondering if I should use it for  `get_inference_helper` too?

python/hsfs/engine/spark.py
vatj: "This behaviour was changed during the implementation of Polars. And dataframe were directly read as Pandas/Polars and Spark dataframe based on the dataframe_type."
It seems like it could break a lot of things. You mean that now you do things that were being done with spark datafra me directly using polars or pandas dataframe?

python/hsfs/core/vector_server.py
```diff
@@ -373,7 +373,7 @@ def get_inference_helper(self, entry, return_type):
         elif return_type.lower() == "polars":
             # Polar considers one dimensional list as a single columns so passed vector as 2d list
             polars_df = pl.DataFrame(
-                [serving_vector], schema=self._feature_vector_col_name, orient="row"
+                [serving_vector], schema=serving_vector.keys(), orient="row"
```
vatj, line 376: Actually can serving vector be None?

vatj (approved):

PR #1261, opened 2024-03-28
[FSTORE-1333] Refactor SQL client out of vector server

vatj:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


python/hsfs/core/transformation_function_engine.py
```diff
@@ -296,3 +312,94 @@ def populate_builtin_transformation_functions(
...
+    ) -> Dict[
+        str, "hsfs.transformation_function_attached.TransformationFunctionAttached"
+    ]:
+        if self._feature_view_api is None:
+            self._feature_view_api = feature_view_api.FeatureViewApi(
+                self._feature_store_id
+            )
+        if self._statistics_engine is None:
+            self._statistics_engine = statistics_engine.StatisticsEngine(
+                self._feature_store_id
```
dhananjay-mk, line 331: entity argument is missing here
TypeError: StatisticsEngine.__init__() missing 1 required positional argument: 'entity_type'

python/hsfs/core/online_store_sql_client.py
```diff
@@ -0,0 +1,718 @@
...
+import logging
+import re
+from typing import Any, Dict, List, Optional, Set, Tuple, Union
+
+from hsfs import feature_view, training_dataset, util
+from hsfs.constructor import serving_prepared_statement
+from hsfs.core import feature_view_api, storage_connector_api, training_dataset_api
+from hsfs.serving_key import ServingKey
+from sqlalchemy import bindparam, exc, sql, text
+from typeguard import typechecked
```
dhananjay-mk, line 27: typeguard is not installed

python/hsfs/core/vector_server.py
```diff
@@ -13,319 +13,212 @@
...
 
     def init_serving(
         self,
-        entity,
-        batch,
-        external,
-        inference_helper_columns=False,
-        options=None,
+        entity: Union["feature_view.FeatureView", "training_dataset.TrainingDataset"],
+        external: bool = None,
```
dhananjay-mk, line 89: default external to False as its bool

dhananjay-mk: also in feature_view.py  `init_serving` method

python/hsfs/core/vector_server.py
```diff
@@ -13,319 +13,212 @@
...
+        inference_helper_columns: bool = False,
+        init_online_store_sql_client: bool = True,
+        init_online_store_rest_client: bool = False,
+        options: Optional[Dict[str, Any]] = None,
     ):
+        self._set_default_online_store_client(
```
dhananjay-mk, line 95: method missing? 
AttributeError: 'VectorServer' object has no attribute '_set_default_online_store_client'

dhananjay-mk (commented):
some comments after trying to run it locally

python/hsfs/core/vector_server.py
```diff
@@ -15,319 +15,196 @@
...
-        batch,
-        external,
-        inference_helper_columns=False,
-        options=None,
+        entity: Union[feature_view.FeatureView, training_dataset.TrainingDataset],
+        external: bool = False,
+        inference_helper_columns: bool = False,
+        options: Optional[Dict[str, Any]] = None,
     ):
         if external is None:
```
dhananjay-mk, line 93: since we made external as bool in argument, this check does not work as the if is never satisfied. So if the user did not specify `external` argument in `init_serving` it will default to False, even if using an external client. I am not sure why we had this check but currently we can remove this code as we expect the user to specify the external flag always. Or we just remove the argument from the signature and always derive the external flag using the client.

vatj: No we should not do that, some people run weird stuff on their cluster. I will fix the check

vatj: Nice job seeing it

dhananjay-mk (approved):
looks good

PR #1263, opened 2024-04-02
[HWORKS-1104] Explicit storage connector provenance

bubriks:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


python/hsfs/core/explicit_provenance.py
```diff
@@ -292,13 +314,13 @@ def from_response_json(
...
-        """Parse explicit links from json response. There are three types of
-        Links: UpstreamFeatureGroups, DownstreamFeatureGroups, DownstreamFeatureViews
+        """Parse explicit links from json response. There are four types of
+        Links: UpstreamStorageConnectors, UpstreamFeatureGroups, DownstreamFeatureGroups, DownstreamFeatureViews
 
         # Arguments
             json_dict: json response from the explicit provenance endpoint
             direction: subset of links to parse - UPSTREAM/DOWNSTREAM
-            artifact: subset of links to parse - FEATURE_GROUP/FEATURE_VIEW/MODEL
+            artifact: subset of links to parse - FEATURE_GROUP/FEATURE_VIEW/MODEL/STORAGE_CONNECTOR
```
robzor92, line 323:
```suggestion
            artifact: subset of links to parse - STORAGE_CONNECTOR/FEATURE_GROUP/FEATURE_VIEW/MODEL
```

python/hsfs/core/feature_group_base_engine.py
```diff
@@ -61,6 +61,22 @@ def get_parent_feature_groups(self, feature_group):
             `ProvenanceLinks`:  the feature groups used to generate this feature group
         """
         return self._feature_group_api.get_parent_feature_groups(feature_group)
+    
+    def get_parent_storage_connectors(self, feature_group):
```
robzor92, line 65: This is mainly for my own understanding, can a feature group have multiple storage connector parents?

robzor92 (changes requested):

python/hsfs/core/feature_group_api.py
```diff
@@ -302,6 +302,43 @@ def get_parent_feature_groups(self, feature_group_instance):
             explicit_provenance.Links.Direction.UPSTREAM,
             explicit_provenance.Links.Type.FEATURE_GROUP,
         )
+    
+    def get_parent_storage_connector(self, feature_group_instance):
```
o-alex, line 306: Jim wanted short method names.
Parent/child are mentioned in the API when you really have to. A fg can have a parent and a child fg.
But a fg can only have a parent storage connector.
I think can reduce this to `get_storage_connector`

bubriks: OK, but now thinking about it are these hsfs changes (`get_storage_connector`) even necessary only on demand (external) fg will have storage connector and when you get it you get the associated storage connector as well.

python/hsfs/feature_group.py
```diff
@@ -512,6 +512,22 @@ def get_parent_feature_groups(self) -> explicit_provenance.Links:
             `hsfs.client.exceptions.RestAPIError`.
         """
         return self._feature_group_engine.get_parent_feature_groups(self)
+    
+    def get_storage_connector(self):
```
o-alex, line 516: ok, had a talk with Jim on this, and he wants things as slim as possible and usable for the user.
I had to do the same for the model provenance.
Can we rename this method to `get_storage_connector_provenance` - this returns a Links object, a object that typical users will not be used to, only powerusers.
Can we create another wrapper method `get_storage_connector` where you extract the storage connector from `accessible`. Similar to: https://github.com/logicalclocks/feature-store-api/blob/master/python/hsfs/feature_view.py

For upstream links of this type, we know a fg can have at most one storage connector.
Maybe you can also check if accessible is empty, if there is something in delete or innaccesible and log a warning if there exists and just return None. I am planning to do similar for models in a future PR.

python/hsfs/storage_connector.py
```diff
@@ -180,6 +180,22 @@ def connector_options(self) -> Dict[str, Any]:
         """
         return {}
 
+    def get_generated_feature_groups(self):
```
o-alex, line 183: similar to previous comment.
can we rename this to `get_feature_groups_provenance` or `get_generated_feature_groups_provenance` - I would argue in this case `generated` same as `child` is not really needed, you need this when you can have child and generated of same type (such as with fgs) or when the name clashes with something else.
And also add a wrapper named `get_feature_groups` that return accessible and if something in deleted or innaccessible, maybe log this, so it is available in job logs/notebook (info)

o-alex (approved):

o-alex (approved):

PR #1275, opened 2024-04-08
[FSTORE-1346] Update dependencies on the documentation generation

vatj:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


vatj:
Do not merge until the keras autodoc dependency has been changed to logicalclocks repo, waiting for corresponding changes to be approved

.github/workflows/mkdocs-master.yml
```diff
@@ -7,23 +7,25 @@ jobs:
...
         with:
-          python-version: '3.8'
-      - run: pip3 install pip==22.0.3
+          python-version: "3.10"
 
       - name: install deps
         working-directory: ./python
         run: cp ../README.md . && pip3 install -e .[python,dev,docs]
 
+      - name: use dev mike
```
robzor92, line 26: I don't remember why we are pinning mike==1.1.2 in setup.py but then installing the latest dev version in the workflow. But it might be more reliable to instead pin mike to 2.0.0 in setup.py and not install mike from dev. Since a bug may leak out in dev causing issues for us.

docs/CONTRIBUTING.md
```diff
@@ -69,42 +67,42 @@ We follow a few best practices for writing the Python documentation:
...
-    `mike` builds the documentation and commits it as a new directory to the gh-pages branch. Each directory corresponds to one version of the documentation. Additionally, `mike` maintains a json in the root of gh-pages with the mappings of versions/aliases for each of the directories available. With aliases you can define extra names like `dev` or `latest`, to indicate stable and unstable releases.
+`mike` builds the documentation and commits it as a new directory to the gh-pages branch. Each directory corresponds to one version of the documentation. Additionally, `mike` maintains a json in the root of gh-pages with the mappings of versions/aliases for each of the directories available. With aliases you can define extra names like `dev` or `latest`, to indicate stable and unstable releases.
 
 1. Currently we are using our own version of `keras-autodoc`
 
-    ```bash
-    pip install git+https://github.com/moritzmeister/keras-autodoc@split-tags-properties
-    ```
+   ```bash
+   pip install git+https://github.com/moritzmeister/keras-autodoc@split-tags-properties
```
robzor92, line 75: Fix this link

docs/overrides/partials/tabs-item.html
```diff
@@ -1,13 +1,16 @@
 <!--
```
robzor92, line 1: What's the reason for changing this file?

robzor92 (changes requested):

.github/workflows/mkdocs-master.yml
vatj: Ok, I think the issue currently is that pyparsing is not compatible with great-expectations but as long as regular user don't install docs I don't think it is an issue as you say. Will pin mike 2.0

docs/overrides/partials/tabs-item.html
vatj: Not compatible with the way partials are overriden when using material theme 9.X.X. It threw some error with nav items. But it is a very clear change in terms of UI if it breaks, so it is easy to spot that it did not break things I would say

robzor92 (approved):

PR #1326, opened 2024-05-07
[FSTORE-1333][APPEND] fg_label and helper fix

vatj:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/core/online_store_sql_client.py
+++ b/python/hsfs/core/online_store_sql_client.py
@@ -42,11 +42,12 @@ class OnlineStoreSqlClient:
         feature_store_id: id,
         skip_fg_ids: Optional[Set[int]],
         serving_keys: Optional[Set[ServingKey]] = None,
+        external: bool = True,
     ):
         _logger.info("Initialising OnlineStoreSqlClient")
         self._feature_store_id = feature_store_id
         self._skip_fg_ids: Set[int] = skip_fg_ids or set()
-        self._external = True
+        self._external = external
 
         self._prefix_by_serving_index = None
         self._pkname_by_serving_index = None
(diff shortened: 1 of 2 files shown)


python/hsfs/core/online_store_sql_client.py
```diff
@@ -42,11 +42,12 @@ def __init__(
         feature_store_id: id,
         skip_fg_ids: Optional[Set[int]],
         serving_keys: Optional[Set[ServingKey]] = None,
+        external: bool = True,
```
dhananjay-mk, line 45: why is the default True, should it be False so that it defaults to internal client?

python/hsfs/core/online_store_sql_client.py
```diff
@@ -42,11 +42,12 @@ def __init__(
...
     ):
-        _logger.info("Initialising OnlineStoreSqlClient")
+        _logger.info("Initialising Online Store Sql Client")
```
dhananjay-mk, line 47: not strictly necessary, may be also here change the log level to debug?

python/hsfs/core/online_store_sql_client.py
```diff
@@ -668,7 +669,9 @@ def serving_keys(self) -> Set[ServingKey]:
                 "Build serving keys from prepared statements ignoring prefix to ensure compatibility with older version."
             )
             self._serving_keys = util.build_serving_keys_from_prepared_statements(
-                self.prepared_statements[self.SINGLE_VECTOR_KEY],
+                self.prepared_statements[
+                    self.BATCH_VECTOR_KEY
```
dhananjay-mk, line 673: not clear why we changed the key from SINGLE to BATCH here?

dhananjay-mk (commented):
minor comments, looks good otherwise

python/hsfs/core/online_store_sql_client.py
vatj: When there is a label fg, batch vector retrieval uses the fg_label prepared statement where as for single vector it does not. So when building keys/prefix etc... we miss the label prepared statement if we use the single vector. As to exactly the current implementation does that, I just don't exactly know

python/hsfs/core/online_store_sql_client.py
vatj: We can although I think it is good to have as otherwise in a notebook one wonders what is happening

python/hsfs/core/online_store_sql_client.py
vatj: Why would we default to the internal client? Anyway the default value is not used as far as I can tell (it is resolved at the Fv level in the init), I can switch the order to make it explicit that it has to be passed.

PR #1344, opened 2024-06-10
[FSTORE-1423] Refactor Great Expectations to make it an optional dependency

vatj:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


python/hsfs/validation_report.py
```diff
@@ -107,8 +111,10 @@ def to_json_dict(self) -> Dict[str, Any]:
             "meta": self._meta,
         }
 
-    def to_ge_type(self) -> ge.core.ExpectationSuiteValidationResult:
-        return ge.core.ExpectationSuiteValidationResult(
+    def to_ge_type(self) -> great_expectations.core.ExpectationSuiteValidationResult:
+        if HAS_GREAT_EXPECTATIONS is False:
```
aversey, line 115: It is better to use `if not HAS_GREAT_EXPECTATIONS:`.

python/hsfs/ge_validation_result.py
```diff
@@ -101,8 +105,10 @@ def to_json_dict(self) -> Dict[str, Any]:
             "meta": self._meta,
         }
 
-    def to_ge_type(self) -> ge.core.ExpectationValidationResult:
-        return ge.core.ExpectationValidationResult(
+    def to_ge_type(self) -> great_expectations.core.ExpectationValidationResult:
+        if HAS_GREAT_EXPECTATIONS is False:
```
aversey, line 109: It is better to use `if not HAS_GREAT_EXPECTATIONS:`.

python/hsfs/core/validation_report_engine.py
```diff
@@ -15,14 +15,20 @@
...
-import great_expectations as ge
 from hsfs import client, util
 from hsfs.core import validation_report_api
 from hsfs.validation_report import ValidationReport
 
 
+HAS_GREAT_EXPECTATIONS = False
+if importlib.util.find_spec("great_expectations") or TYPE_CHECKING:
+    HAS_GREAT_EXPECTATIONS = True
+    import great_expectations
```
aversey, line 29: This is strange. If we don't have GE, but have TYPE_CHECKING, it will set HAS_GREAT_EXPECTATIONS to True (but then fail on import). It seems that `or TYPE_CHECKING` should be removed.

I propose to add `HAS_GREAT_EXPECTATIONS = importlib.util.find_spec("great_expectations") is not None` to core.constants instead. Also, maybe an overkill, but it can make sense to create a decorator:
```
def uses_great_expectations(f):
    @functools.wraps(f)
    def g(*args, **kwds):
        if not HAS_GREAT_EXPECTATIONS:
            raise ModuleNotFoundError(great_expectations_not_installed_message)
        return f(*args, **kwds)
    return g
```

aversey (commented):
Overall looks good to me, I only have minor remarks.

python/hsfs/core/validation_report_engine.py
vatj: Regarding the "or TYPE_CHECKING", TYPE_CHECKING will be true everywhere else so it will fail in every file we have the check. Why are you concern about it failing here?

Decorator is a good idea, maybe something we should add systematically for "special dependencies". We can start with GE, see how it works out for now.

aversey: Ok, I have thought more about this and I think it is fine to use `or TYPE_CHECKING` here. I initially thought that it is completely redundant, but it actually changes the error type checker shows (import vs use), which I guess is a bit better.

python/tests/core/test_great_expectation_engine.py
```diff
@@ -13,13 +13,21 @@
...
+import pandas as pd
+import pytest
 from hsfs import feature_group, validation_report
 from hsfs.core import great_expectation_engine
 
 
+HAS_GREAT_EXPECTATIONS = False
+if importlib.util.find_spec("great_expectations"):
+    HAS_GREAT_EXPECTATIONS = True
+    import great_expectations.core
```
aversey, line 28: Is there a reason not to use hsfs.core.constans.HAS_GREAT_EXPECTATIONS here?

aversey (approved):

python/tests/core/test_great_expectation_engine.py
vatj: Ah nej sorry, forgot to extend the search to the tests folders

PR #1354, opened 2024-07-02
[FSTORE-1424] Feature logging

kennethmhc:
This PR adds/fixes/changes...

JIRA Issue: -

Priority for Review: -

Related PRs: -

**How Has This Been Tested?**

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Tests on VM

```
```


python/hsfs/core/feature_view_api.py
```diff
@@ -358,3 +366,98 @@ def get_models_provenance(
...
+        _client = client.get_instance()
+        path_params = self._base_path + [
+            feature_view_name,
+            self._VERSION,
+            feature_view_version,
+            self._LOGGING,
+        ]
+        return _client._send_request("GET", path_params, {})
+
+    def delete_feature_log(
```
javierdlrm, line 445: Not sure what this method is for.
Is it deleting the features that have been logged before or just disabling feature logging?
If it's disabling feature logging only, I would call it `disable_feature_logging` since we have `enable_feature_logging`.
If it's deleting the features previously logged, I would call it `delete_feature_logging`.

python/hsfs/core/feature_view_engine.py
```diff
@@ -923,3 +935,138 @@ def _check_if_exists_with_prefix(self, f_name, f_set):
             )
         else:
             return f_name
+
+    def enable_feature_logging(self, fv):
+        self._feature_view_api.enable_feature_logging(fv.name, fv.version)
+        fv.enabled_logging = self.get(fv.name, fv.version).enabled_logging
```
javierdlrm, line 941: Why do we need two requests to the backend? I would assume that if the call to `enable_feature_logging` doesn't raise an error, it went through and then we can just set `fv.enabled_logging` to true.

python/hsfs/core/feature_logging.py
```diff
@@ -0,0 +1,51 @@
...
+import humps
+from hsfs import util
+
+
+class FeatureLogging:
+
+    def __init__(self, id, transformed_features, untransformed_features):
+        self._id = id
+        self._transformed_features = transformed_features
+        self._untransformed_features = untransformed_features
```
javierdlrm, line 14: It took me a while to realize that these are actually Feature Group objects.
Can we rename them to something like `transformed_feature_group` | `untransformed_feature_group` or `transformed_features_fg` | `untransformed_features_fg` ?

python/hsfs/core/feature_view_engine.py
```diff
@@ -923,3 +935,138 @@ def _check_if_exists_with_prefix(self, f_name, f_set):
...
+            [feature for feature in fv.features if not feature.label],
+            [feature for feature in fv.features if feature.label],
+            FeatureViewEngine._LOG_TD_VERSION,
+            FeatureViewEngine._LOG_TIME,
+            FeatureViewEngine._HSML_MODEL,
+            prediction,
+            training_dataset_version,
+            hsml_model,
+        )
+        return fg.insert(df, write_options=default_write_options)
```
javierdlrm, line 974: Have you thought of using the `multi_insert` method to send feature vectors directly to kafka without batching them first?

python/hsfs/core/feature_view_engine.py
```diff
@@ -923,3 +935,138 @@ def _check_if_exists_with_prefix(self, f_name, f_set):
...
+            query = query.filter(fg.get_feature(FeatureViewEngine._LOG_TIME) <= end_time)
+        if training_dataset_version:
+            query = query.filter(fg.get_feature(FeatureViewEngine._LOG_TD_VERSION) == training_dataset_version)
+        if hsml_model:
+            query = query.filter(fg.get_feature(FeatureViewEngine._HSML_MODEL) == self.get_hsml_model_value(hsml_model))
+        if filter:
+            query = query.filter(self._convert_to_log_fg_filter(fg, fv, filter, fv_feat_name_map))
+        df = query.read()
+        df = df.drop(["log_id", FeatureViewEngine._LOG_TIME], axis=1)
+        return df
```
javierdlrm, line 1001: Have you thought of removing this method and make the users get the features directly from the feature groups? They can read the features via `fv.feature_logging.transformed_feature_group.read()`.
It would simplify our API.

python/hsfs/feature_group.py
```diff
@@ -140,6 +140,7 @@ def __init__(
         self._notification_topic_name = notification_topic_name
         self._deprecated = deprecated
         self._feature_store_id = featurestore_id
+        self._feature_store = None
```
javierdlrm, line 143: Do we need this parameter? I think we are not using it.

python/hsfs/feature_view.py
```diff
@@ -3437,11 +3446,218 @@ def update_from_response_json(self, json_dict: Dict[str, Any]) -> "FeatureView":
...
+
+        Note: If features is a `pd.Dataframe`, prediction can be provided as columns in the dataframe.
+
+        # Arguments
+            features: The features to be logged. Can be a pandas DataFrame, a list of lists, or a numpy ndarray.
+            prediction: The predictions to be logged. Can be a pandas DataFrame, a list of lists, or a numpy ndarray. Defaults to None.
+            transformed: Whether the features are transformed. Defaults to False.
+            write_options: Options for writing the log. Defaults to None.
+            training_dataset_version: Version of the training dataset. Defaults to None.
+            hsml_model: hsml.model.Model` HSML model associated with the log. Defaults to None.
```
javierdlrm, line 3506: Is a backtick ( ` ) missing here?

python/hsfs/feature_view.py
```diff
@@ -3437,11 +3446,218 @@ def update_from_response_json(self, json_dict: Dict[str, Any]) -> "FeatureView":
...
+            # enable logging
+            feature_view.enable_logging()
+            ```
+
+        # Raises
+            `hsfs.client.exceptions.RestAPIError` in case the backend fails to enable feature logging.
+        """
+        return self._feature_view_engine.enable_feature_logging(self)
+
+    def log(self,
```
javierdlrm, line 3488: There is a limitation with these arguments. It's not possible to log both untransformed and transformed features at the same time. That would be possible if we reorganize the parameters as
`
def log(self,
    transformed_features,
    untransformed_features,
    predictions,
    ...
`

python/hsfs/feature_view.py
```diff
@@ -3437,11 +3446,218 @@ def update_from_response_json(self, json_dict: Dict[str, Any]) -> "FeatureView":
...
+            features: Union[pd.DataFrame, list[list], np.ndarray],
+            prediction: Optional[Union[pd.DataFrame, list[list], np.ndarray]]=None,
```
javierdlrm, line 3490: `predictions` seems more accurate.

javierdlrm (changes requested):

python/hsfs/feature_view.py
```diff
@@ -3437,11 +3446,218 @@ def update_from_response_json(self, json_dict: Dict[str, Any]) -> "FeatureView":
...
+            features, self.transformation_functions
+        )
+
+    def transform_feature_vector(self, features):
+        return self.transform_feature_vectors([features])
+
+    def transform_feature_vectors(self, features):
+        return self._batch_scoring_server.transform_feature_vectors(features)
+
+    def enable_logging(self) -> None:
```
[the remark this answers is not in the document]
kennethmhc: @jimdowling prefer the shorter form `enable_logging`

python/hsfs/core/feature_logging.py
kennethmhc: added the type hints instead to make the variable name more concise.

python/hsfs/core/feature_view_api.py
kennethmhc: it means deleting the features previously logged, I modified the method description. It is named similar to `read_log`

python/hsfs/core/feature_view_engine.py
kennethmhc: It is to make sure the backend enable logging successfully. Since this method `enable_feature_logging` will only be called once, the impact on the performance is small. But maybe it should check `enabled_logging=True` here.

python/hsfs/core/feature_view_engine.py
kennethmhc: The problem is  it require users to call `finalize_multi_part_insert` otherwise the log wont be send to kafka

python/hsfs/core/feature_view_engine.py
kennethmhc: I think it is better to hide the implementation detail as much as possible so that it is easier for us to make changes in the future. Also this method provide an easy way for user to filter log by time, td version and model version.

python/hsfs/feature_group.py
kennethmhc: It is required, otherwise `fg.feature_store` will fail if setter is not called before.

python/hsfs/feature_store.py
```diff
@@ -1592,6 +1594,7 @@ def get_or_create_feature_view(
         inference_helper_columns: Optional[List[str]] = None,
         training_helper_columns: Optional[List[str]] = None,
         transformation_functions: Optional[Dict[str, TransformationFunction]] = None,
+        enabled_logging: bool = False,
```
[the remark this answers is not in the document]
kennethmhc: changed it to `logging_enabled`

python/hsfs/feature_view.py
kennethmhc: good point, may be we can change it to 
```
def log(self,
            transformed_features: Optional[Union[pd.DataFrame, list[list], np.ndarray]]=None,
            untransformed_features: Optional[Union[pd.DataFrame, list[list], np.ndarray]]=None,
            prediction: Optional[Union[pd.DataFrame, list[list], np.ndarray]]=None,
            write_options: Optional[Dict[str, Any]] = None,
            training_dataset_version: Optional[int]=None,
            hsml_model=None,
            )
```
@jimdowling agree?

python/hsfs/core/feature_view_engine.py
javierdlrm: Alright! that's fine.
What about renaming the method to `read_feature_logs`?

javierdlrm (changes requested):

python/hsfs/core/feature_view_api.py
kennethmhc: rename the method in `feature_view_engine` for the name in `feature_view` I think Jim prefer not including the `feature` in the name.

python/hsfs/core/feature_view_engine.py
kennethmhc: We can extend this with the `multi_insert` as an extra option later. It would be nice to have if users do not forget to call `finalize_multi_part_insert`. But if there is a container to flush the feature, then it should be fine.

python/hsfs/core/feature_view_engine.py
kennethmhc: renamed

python/hsfs/feature_view.py
```diff
@@ -3437,11 +3450,218 @@ def update_from_response_json(self, json_dict: Dict[str, Any]) -> "FeatureView":
...
+            feature_view.log(features)
+            # log features and predictions
+            feature_view.log(features, prediction)
+            ```
+
+        # Raises
+            `hsfs.client.exceptions.RestAPIError` in case the backend fails to log features.
+        """
+        if not self.logging_enabled:
+            self.enable_logging()
```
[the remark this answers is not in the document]
kennethmhc: I think we can throw a warning here instead of exception because it gives a better user experience.

python/hsfs/core/feature_view_engine.py
javierdlrm: It was renamed in `feature_view_engine`, but not in `feature_view` :) Do you want to rename it there as well?

javierdlrm (changes requested):

python/hsfs/core/feature_view_engine.py
javierdlrm: Could it be that the response is 200 but feature logging was not enabled successfully for that feature view? (e.g., if creating one of the feature group fails).

Generally, if enabling feature logging fails in the backend, an exception should be raise and the response of the request would be != 200. So, a 200 response is "verifying" that feature logging was enabled successfully.
If logging_enabled was an object, another approach would be that the backend returns the updated object in the response and hsfs update feature view inside the same method in feature_view_api. But in this case is a boolean, so 200 response would suffice I believe :)

javierdlrm (changes requested):

python/hsfs/feature_store.py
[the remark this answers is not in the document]
javierdlrm: This description is still missing :)

python/hsfs/core/feature_view_engine.py
kennethmhc: ok, set it to True

python/hsfs/core/feature_view_engine.py
kennethmhc: Let change the name later after Jim come back

python/hsfs/feature_store.py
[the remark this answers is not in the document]
kennethmhc: I added the description `logging_enabled: If true, enable feature logging for the feature view.`

javierdlrm (approved):

PR #1371, opened 2024-07-10
[FSTORE-1411] On-Demand Transformation Functions

manu-sj:
This PR adds functionality for On-demand transformation function as per the design log https://hopsworks.atlassian.net/wiki/spaces/FST/pages/461733903/On-Demand+Transformation+Functions+-+User+API+and+Implementation

Changes Done:

1. Allow custom UDF's to drop features after transformations. The feature is implemented by adding a parameter drop in the decorator.
```
@hopsworks.udf(return_type=float, drop=["arg1", "arg2"])
def custom_udf(arg1, arg2, arg3):
    ...
    return value
```
2. Removed `TransformationFunctionsAttached` from the python client 
3. Added ability to attach transformation functions to feature groups. (On-Demand Transformations)
4. Added to feature view ability to recompute on-demand features using request parameters.
5. Added to feature view  ability to manually compute on-demand features using the function `compute_on_demand_features `
6. Added to feature view ability to access manually apply model-dependent transformations to the feature vector `transform`.
7. Added the ability to access executable model-dependent transformation function using a dictionary with the key was the output column name. (Only output column name is unique for a model dependent transformation function).
8. Added the ability to access executable on-demand transformation function using a dictionary with the key was the on-demand feature name. (Only output column name is unique for a model dependent transformation function).

JIRA Issue: https://hopsworks.atlassian.net/browse/FSTORE-1411

Priority for Review: -

Related PRs:
https://github.com/logicalclocks/hopsworks-ee/pull/1862
https://github.com/logicalclocks/hopsworks-chef/pull/1148
https://github.com/logicalclocks/loadtest/pull/394

**How Has This Been Tested?**

- [x] Unit Tests
- [X] Integration Tests
- [x] Manual Tests on VM

```
```


python/hsfs/hopsworks_udf.py
```diff
@@ -127,11 +145,21 @@ def __init__(
         return_types: Union[List[type], type, List[str], str],
         name: Optional[str] = None,
         transformation_features: Optional[List[TransformationFeature]] = None,
+        transformation_function_argument_names: Optional[
+            List[TransformationFeature]
+        ] = None,
+        dropped_argument_names: Optional[List[str]] = None,
+        dropped_feature_names: Optional[List[str]] = None,
+        feature_name_prefix: Optional[str] = None,
```
kennethmhc, line 153: this is needed for MDT only right?

python/hsfs/hopsworks_udf.py
```diff
@@ -102,11 +117,14 @@ class HopsworksUdf:
 
     Attributes
     ----------
-        output_type (List[str]) : Output types of the columns returned from the UDF.
         function_name (str) : Name of the UDF
-        statistics_required (bool) : True if statistics is required for any of the parameters of the UDF.
-        transformation_statistics (Dict[str, FeatureDescriptiveStatistics]): Dictionary that maps the statistics_argument name in the function to the actual statistics variable.
+        udf_type (UDFType): Type of the UDF can be either \"model dependent\" or \"on-demand\".
```
kennethmhc, line 121: not available in the constructor

python/hsfs/hopsworks_udf.py
```diff
@@ -519,6 +629,8 @@ def get_udf(self, force_python_udf: bool = False) -> Callable:
         # Returns
             `Callable`: Pandas UDF in the spark engine otherwise returns a python function for the UDF.
         """
+        if self.udf_type is None:
```
kennethmhc, line 632: why is it needed?

python/hsfs/hopsworks_udf.py
```diff
@@ -31,7 +32,18 @@
 from hsfs.transformation_statistics import TransformationStatistics
 
 
-def udf(return_type: Union[List[type], type]) -> "HopsworksUdf":
+class UDFType(Enum):
```
kennethmhc, line 35: This is actually transformation type. The UDF is actually the same and can be used for both ODT and MDT. I understand this type is needed for the udf to return different column names, but imo this is better to be done in `transformation_function` class and also the validation of the transformation type. But it is not critical, can be refactored later.

python/hsfs/feature_view.py
```diff
@@ -3486,16 +3513,62 @@ def update_from_response_json(self, json_dict: Dict[str, Any]) -> "FeatureView":
...
-    def transform_batch_data(self, features):
-        return self._feature_view_engine.transform_batch_data(
-            features, self._batch_scoring_server._transformation_functions
+    def compute_on_demand_features(
+        self,
+        feature_vector: Union[List[Any], List[List[Any]], pd.DataFrame, pl.DataFrame],
+        request_parameters: Optional[
+            Union[List[Dict[str, Any]], Dict[str, Any]]
+        ] = None,
+        external: Optional[bool] = None,
```
kennethmhc, line 3522: why is it needed? this function won't fetch features from online store right?

python/hsfs/feature_view.py
```diff
@@ -3486,16 +3513,62 @@ def update_from_response_json(self, json_dict: Dict[str, Any]) -> "FeatureView":
...
+        return self._batch_scoring_server.compute_on_demand_features(
+            feature_vectors=feature_vector, request_parameters=request_parameters
         )
 
-    def transform_feature_vector(self, features):
-        return self.transform_feature_vectors([features])
+    def transform(
+        self,
+        feature_vector: Union[List[Any], List[List[Any]], pd.DataFrame, pl.DataFrame],
+        external: Optional[bool] = None,
```
kennethmhc, line 3549: why is it needed? this function won't fetch features from online store right?

kennethmhc (commented):
Comments about the user facing API.

python/hsfs/feature_group.py
```diff
@@ -2189,6 +2194,47 @@ def __init__(
         self._feature_writers: Optional[Dict[str, callable]] = None
         self._writer: Optional[callable] = None
         self._kafka_headers: Optional[Dict[str, bytes]] = None
+        # On-Demand Transformation Functions
+        self._transformation_functions: List[TransformationFunction] = []
+
+        if transformation_functions:
```
kennethmhc, line 2200: need to consider transformation in `feature_group.save` as well, when feature are defined in `create_feature_group`.

kennethmhc (changes requested):

python/hsfs/engine/python.py
```diff
@@ -920,7 +944,7 @@ def get_training_data(
             #        training_dataset_obj, feature_view_obj, training_dataset_version
             #    )
             return self._apply_transformation_function(
-                training_dataset_obj.transformation_functions, df
+                feature_view_obj.transformation_functions, df
```
kennethmhc, line 947: `feature_view_obj.transformation_functions` has both ODT and MDT?

python/hsfs/feature_view.py
```diff
@@ -1033,7 +1055,7 @@ def get_batch_data(
             start_time,
             end_time,
             self._batch_scoring_server.training_dataset_version,
-            self._batch_scoring_server._transformation_functions,
+            self._batch_scoring_server._model_dependent_transformation_functions,
```
kennethmhc, line 1058: need to include ODT as well?

python/hsfs/hopsworks_udf.py
manu-sj: This is also required for on-demand features. 

So if the query used while creating a feature view uses prefix then all features from that feature group in the feature view would be prefixed with the name. So we need to add the prefix in the new output feature generated and also when trying to acess a specific feature for providing as input to the transformations function.

`query = fg1.select_all().join(fg2.select_all(), prefix="fg2_")`

python/hsfs/hopsworks_udf.py
```diff
@@ -102,11 +117,14 @@ class HopsworksUdf:
...
     ----------
-        output_type (List[str]) : Output types of the columns returned from the UDF.
         function_name (str) : Name of the UDF
-        statistics_required (bool) : True if statistics is required for any of the parameters of the UDF.
-        transformation_statistics (Dict[str, FeatureDescriptiveStatistics]): Dictionary that maps the statistics_argument name in the function to the actual statistics variable.
+        udf_type (UDFType): Type of the UDF can be either \"model dependent\" or \"on-demand\".
+        return_types (List[str]): The data types of the columns returned from the UDF.
         transformation_features (List[str]) : List of feature names to which the transformation function would be applied.
+        output_column_names (List[str]): Column names of the DataFrame returned after application of the transformation function.
+        dropped_features (List[str]): List of features that will be dropped after the UDF is applied.
```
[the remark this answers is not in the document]
manu-sj: The whole Attribute list was based on the accessable attributes present in the class create based on the `@property` decorator. Should I change it to reflect the input arguments to the class.

python/hsfs/hopsworks_udf.py
manu-sj: You are right. I will move it out to transformation function class.

python/hsfs/engine/python.py
manu-sj: No it just has MDT. ODT are only required to be run during online inference. While creation of training data and batch data. The on-demand features are read from the feature group directly without any recomputation.

python/hsfs/hopsworks_udf.py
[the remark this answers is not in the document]
kennethmhc: ok. maybe add description for the arguments

python/hsfs/feature_view.py
manu-sj: While getting batch data the on-demand featuers are not re-computed but read directly from the feature group. So it is not actually required to pass ODT's.

python/hsfs/feature_view.py
manu-sj: Yes you are right it won't fetch features from the online store, but MDT's might require statistics of features. The statistics used while transforming depends on the training_dataset_version set in `init_serving` so if `init_serving` had not been done we don't know the statistics to be used for applying the transformation function. So I basically copied the code done in `get_feature_vector` for initializing vector server. 

I could be wrong here but I think there should be some check to see if `init_serving` or `init_batch_scoring` has been performed.

python/hsfs/core/vector_server.py
```diff
@@ -451,17 +548,174 @@ def assemble_feature_vector(
...
+            return_type = "pandas"
+            feature_vectors = feature_vectors.to_dict(orient="records")
+
+        elif isinstance(feature_vectors, pl.DataFrame):
+            return_type = "polars"
+            feature_vectors = feature_vectors.to_pandas()
+            feature_vectors = feature_vectors.to_dict(orient="records")
+
+        elif isinstance(feature_vectors, list) and feature_vectors:
+            if isinstance(feature_vectors[0], list):
```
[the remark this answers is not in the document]
manu-sj: Updated check to see if all items are lists

python/hsfs/feature_group.py
manu-sj: Updated to handle cases where create_feature_group is defined with features and feature_group.save if called.

To handle this usecase I moved the function to updating the schema based on transformation functions from `python/spark engine` to `feature_group_engine`. In `feature_group_engine` the `_update_feature_group_schema_on_demand_transformations` function updates the schema to add transformed features and drop any feature if specified in the transformation function.

python/hsfs/hopsworks_udf.py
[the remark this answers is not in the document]
manu-sj: Updated docs to have arguments also

kennethmhc (approved):
LGTM (the rest of the comment will be addressed in a separate PR)

PR #1379, opened 2024-09-05
[FSTORE-1533] Importing hopsworks fails due to core dump while importing polars

manu-sj:
This PR fixes an issue in which hopworks is not importable because Polars throws a "" due to AVX instructions being unavailable.

https://github.com/pola-rs/polars/issues/2922 
https://github.com/pola-rs/polars?tab=readme-ov-file#legacy 

Fix Done: 
Make polars a optional dependency. The changes being done closely resembles the changes done for 4.0 (https://github.com/logicalclocks/feature-store-api/pull/1363)

JIRA Issue: https://hopsworks.atlassian.net/browse/FSTORE-1533

Related PRs: 
https://github.com/logicalclocks/hopsworks-api/pull/318
https://github.com/logicalclocks/loadtest/pull/454

**How Has This Been Tested?**

- [x] Unit Tests
- [ ] Integration Tests
- [x] Manual Tests on VM

```
```

### DIFF
--- a/python/hsfs/core/vector_server.py
+++ b/python/hsfs/core/vector_server.py
@@ -21,7 +21,6 @@ from base64 import b64decode
 from datetime import datetime, timezone
 from io import BytesIO
 from typing import (
-    TYPE_CHECKING,
     Any,
     Callable,
     Dict,
@@ -70,7 +69,7 @@ try:
 except ImportError:
     from avro.io import BinaryDecoder
 
-if HAS_POLARS or TYPE_CHECKING:
+if HAS_POLARS or HAS_POLARS:
     import polars as pl
 
 _logger = logging.getLogger(__name__)
--- a/python/hsfs/core/vector_server.py
+++ b/python/hsfs/core/vector_server.py
@@ -69,7 +69,7 @@ try:
 except ImportError:
     from avro.io import BinaryDecoder
 
-if HAS_POLARS or HAS_POLARS:
+if HAS_POLARS:
     import polars as pl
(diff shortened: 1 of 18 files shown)


python/hsfs/core/vector_server.py
```diff
@@ -59,6 +69,9 @@
 except ImportError:
     from avro.io import BinaryDecoder
 
+if HAS_POLARS or HAS_POLARS:
```
SirOibaf: Typo here.

python/hsfs/engine/python.py
```diff
@@ -206,7 +218,6 @@ def _sql_offline(
         hive_config: Optional[Dict[str, Any]] = None,
         arrow_flight_config: Optional[Dict[str, Any]] = None,
     ) -> Union[pd.DataFrame, pl.DataFrame]:
```
SirOibaf, line 220: What about these annotations? Do they work if polars is not installed?

manu-sj: Yes there are just type hints are they would not cause any run time effects. I have tested it with the unit test and it was working. Still need to run some workflow tests and see if everything is fine.

https://stackoverflow.com/questions/41692473/does-python-type-hint-annotations-cause-some-run-time-effects

python/hsfs/core/vector_server.py
manu-sj: Fixed type

SirOibaf (approved):

